PR #1, opened 2018-06-09
(WIP) Properties and components

brianpeiris:


src/containers/NodePropertyGroupContainer.js
```diff
@@ -1,14 +1,50 @@
...
+      name: node.name
+    });
+  };
+  updateName = e => {
+    if (!this.state.node) return;
+    this.props.editor.execute(new SetValueCommand(this.state.node, "name", e.target.value));
+  };
+  updatePosition = newPosition => {
+    if (!this.state.node) return;
+    this.props.editor.execute(new SetPositionCommand(this.state.node, new THREE.Vector3().copy(newPosition)));
```
brianpeiris: Dunno if it's a bad idea to require and use `THREE` here, but `SetPositionCommand` expects a Vector3.

src/components/Vector3Input.js
```diff
@@ -7,11 +7,11 @@ export default function Vector3Input({ value, onChange, ...rest }) {
   return (
     <div className={styles.vector3Input}>
       <div className={styles.label}>X:</div>
-      <NumericInput value={value.x} onChange={x => onChange("x", x)} {...rest} />
+      <NumericInput value={value.x} onChange={x => onChange({ x, y: value.y, z: value.z })} {...rest} />
```
brianpeiris: Had the `Vector3Input` send the full state since it would be difficult for the receiver to reconstruct the rest of the state otherwise.

src/containers/NodePropertyGroupContainer.js
robertlong: It's probably fine to include `THREE` in the container. I'd try to avoid it any further down the hierarchy.

robertlong: It should be safe to store and set a local THREE.Vector3 every time you update the position. `SetPositionCommand` already clones the new position: https://github.com/MozillaReality/hubs-editor/blob/master/src/editor/commands/SetPositionCommand.js#L28

src/components/Vector3Input.js
```diff
@@ -7,11 +7,11 @@ export default function Vector3Input({ value, onChange, ...rest }) {
...
       <div className={styles.label}>Y:</div>
-      <NumericInput value={value.y} onChange={y => onChange("y", y)} {...rest} />
+      <NumericInput value={value.y} onChange={y => onChange({ x: value.y, y, z: value.z })} {...rest} />
```
robertlong: Check the key value pairs in your calls to onChange. x,y, and z don't all match up.

brianpeiris: Ah, good catch.

PR #4, opened 2018-07-07
Save scene state when breadcrumbing

brianpeiris:
Save scenes instead of replacing them in order to maintain scene state when breadcrumbing.


src/client/editor/Editor.js
```diff
@@ -147,65 +159,42 @@ export default class Editor {
...
-
-    while (this.scene.children.length > 0) {
-      this.removeObject(this.scene.children[0]);
-    }
-
-    while (scene.children.length > 0) {
-      this.addObject(scene.children[0]);
-    }
-
-    this.signals.sceneGraphChanged.active = true;
```
brianpeiris: Not sure if there was any point to replacing children in the scene instead of just using the new scene itself.

PR #37, opened 2018-07-10
Remove button for components

tony056:
- Changes:
  - Added remove button to PropertyGroup.
  - Added RemoveComponentCommand.
  - Pushed the remove button to the right in PropertyGroup.scss

### DIFF
--- a/src/client/components/PropertyGroup.js
+++ b/src/client/components/PropertyGroup.js
@@ -1,11 +1,22 @@
 import React from "react";
+import Button from "./Button";
 import PropTypes from "prop-types";
 import styles from "./PropertyGroup.scss";
 
-export default function PropertyGroup({ name, children }) {
+export default function PropertyGroup({ name, removeHandler, children }) {
+  const displayRemoveButton = (name, removeHandler) => {
+    if (name === "Node") {
+      return;
+    }
+    return <Button onClick={removeHandler}>Remove</Button>;
+  };
+
   return (
     <div className={styles.propertyGroup}>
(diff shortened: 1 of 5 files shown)


src/client/components/PropertyGroup.js
```diff
@@ -1,17 +1,29 @@
 import React from "react";
+import Button from "./Button";
 import PropTypes from "prop-types";
 import styles from "./PropertyGroup.scss";
 
-export default function PropertyGroup({ name, children }) {
+export default function PropertyGroup({ name, removeHandler, children }) {
+  const displayRemoveButton = (name, removeHandler) => {
+    if (name === "Node") {
```
brianpeiris: I would add a `removable` property instead of basing this on `name`.

src/client/editor/commands/RemoveComponentCommand.js
```diff
@@ -0,0 +1,32 @@
...
+  execute() {
+    this.editor.removeComponent(this.object, this.componentName);
+    this.editor.removeHelper(this.object);
+    this.editor.signals.objectChanged.dispatch(this.object);
+  }
+
+  undo() {
+    this.editor.addComponent(this.object, this.componentName);
+    this.object.traverse(child => {
+      this.editor.addHelper(child);
```
brianpeiris: addHelper needs a `selectionRoot` parameter. `this.object` should work.

brianpeiris: Oh, actually, it looks like `addComponent` already takes care of adding helpers.

PR #38, opened 2018-07-10
Menu Bar (Component without functional menu items)

robertlong:


src/client/components/Menu.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
```
brianpeiris: I would put the menu components in a `components/menu/` sub directory.

robertlong: Maybe we can restructure the components in another PR? I don't think this is a special case. Inputs should probably go in a folder as well if we want to do that. Personally I think we can keep the flat structure for now.

PR #40, opened 2018-07-11
Improve color picker

tony056:
- Changes:
  - Added a color block and an input as the default color input.
  - Enabled color picker after clicking on the color block.
  - Remove alpha channel in the color picker.


tony056:
Got it, thanks!

PR #50, opened 2018-07-11
FileInput and file dialog

### LINKED ISSUE #6: File Picker
For picking material textures and scene/gltf references

brianpeiris:
Closes #6


src/client/ui/inputs/FileInput.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import styles from "./FileInput.scss";
+import Button from "../Button";
+
+export default function FileInput({ value, onChange, openFileDialog, filters }) {
+  return (
+    <div className={styles.fileInput}>
+      <input value={value} onChange={e => onChange(e.target.value)} />
```
robertlong: Do we want this input to be editable? Seems like we would only want this to change when you select a file.

brianpeiris: Yeah, I wasn't sure. I like the idea of being able to type in a file path/name directly if you already know it offhand. It does cause a bunch of errors because onChange is apparently fired for each keystroke, but we can fix that part of it.
I'll merge as is and we can decide later?

robertlong: I'd rather remove onChange for that input. It does cause a bunch of errors.

brianpeiris: I'd rather fix the errors, so I'll do that.

robertlong: Ok, can we make sure that the file paths must be valid or null when saving?

brianpeiris: I'll leave that to another PR. Sounds like it's more part of the conflict resolution features.

robertlong: Yes, I also want to ensure that the export step works correctly.

PR #55, opened 2018-07-16
Warning display - duplicate node names in the inherited file.

tony056:
- Changes:
  - Displayed the warning message with `SnackBar` at the bottom of `HierarchyPanel`.
  - Render duplicate nodes with the warning style.
  - Automatically generate node names with the node path if there's a duplicate.
  - Added the tooltip on nodes with duplicate names to show the original name.

### DIFF
--- a/src/client/ui/panels/HierarchyPanelContainer.js
+++ b/src/client/ui/panels/HierarchyPanelContainer.js
@@ -135,16 +135,20 @@ class HierarchyPanelContainer extends Component {
   };
 
   renderNode = node => {
+    const isConflict = node.object.userData._missing || node.object.userData._duplicate;
+    const isMissingChild = node.object.userData._missing && !node.object.userData._isMissingRoot;
+    const isDuplicateChild = node.object.userData._duplicate && !node.object.userData._isDuplicateRoot;
     return (
       <div
         className={classNames("node", {
           "is-active": this.props.editor.selected && node.object.id === this.props.editor.selected.id,
-          conflict: node.object.userData._missing,
+          conflict: isConflict,
           "error-root": node.object.userData._isMissingRoot ? node.object.userData._missing : false,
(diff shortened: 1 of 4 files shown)


src/client/ui/panels/HierarchyPanelContainer.js
```diff
@@ -135,16 +136,20 @@ class HierarchyPanelContainer extends Component {
...
+          conflict: isConflict,
           "error-root": node.object.userData._isMissingRoot ? node.object.userData._missing : false,
-          disabled: node.object.userData._missing && !node.object.userData._isMissingRoot
+          "warning-root": node.object.userData._isDuplicateRoot ? node.object.userData._duplicate : false,
+          disabled: isMissingChild || isDuplicateChild
         })}
-        onMouseUp={node.object.userData._missing ? null : e => this.onMouseUpNode(e, node)}
-        onMouseDown={node.object.userData._missing ? e => e.stopPropagation() : null}
+        onMouseUp={isConflict ? null : e => this.onMouseUpNode(e, node)}
+        onMouseDown={isConflict ? e => e.stopPropagation() : null}
```
brianpeiris: Do we want to disable interaction with duplicate objects?

tony056: I'm not sure about it, so I disable them at first. My concern is that if users move children to other duplicate roots and then save the file. The `.scene` file does not really take care of this situation since we don't modify the node name in the gltf file, which might made users confused when they reopen the `.scene` file and see all children are still under the first duplicate root.

brianpeiris: Fair enough. Please make a new issue for it so that we can deal with it later.

tony056:
I keep both `_duplicate` and `_isDuplicateRoot` now to disable them temporarily.
I will open two new issues, one is to deal with unnamed nodes, another one is to deal with the situation I mentioned above.

PR #59, opened 2018-07-18
Enable orphaned nodes editing

tony056:
- Features:
  - Remove orphaned nodes
  - Move orphaned nodes to other nodes
  - Update conflict resolution
- Add:
  - Included two sample scenes with missing nodes or duplicate node names in the inherited file (`missing-n-duplicate.scene` and `missing-node2.scene`)
orphaned-nodes

### DIFF
--- a/example/missing-n-duplicate.scene
+++ b/example/missing-n-duplicate.scene
@@ -1,0 +1,37 @@
+{
+  "entities": {
+    "New_Node1": {
+      "parent": "Pillar1x3RndShape",
+      "index": 0
+    },
+    "New_Node2": {
+      "parent": "Pillar1x3RndShape",
+      "index": 1
+    },
+    "New_Node3": {
+      "parent": "Pillar1x3Rnd1",
+      "index": 0
+    },
+    "New_Node4": {
+      "parent": "New_Node3",
+      "index": 0
+    },
+    "New_Node5": {
+      "parent": "New_Node4",
+      "index": 0
+    },
+    "New_Node6": {
+      "parent": "New_Node4",
+      "index": 0
+    },
+    "New_Node7": {
+      "parent": "New_Node6",
+      "index": 0
+    },
+    "New_Node8": {
+      "parent": "New_Node6",
+      "index": 0
+    }
+  },
+  "inherits": "./ArchitectureKit/Pillar1x3Rnd.gltf"
+}
--- a/src/client/editor/ConflictResolution.js
+++ b/src/client/editor/ConflictResolution.js
@@ -1,0 +1,97 @@
+export default class ConflictHandler {
+  constructor() {
+    this._conflicts = {
+      missing: false,
+      duplicate: false
+    };
+    this.duplicateList = {};
+  }
+
+  findDuplicates = (node, layer, index) => {
+    if (node.userData._path) {
+      node.userData._path.push(index);
+    } else {
+      node.userData._path = [0];
+    }
+
+    const name = node.name;
+    this.duplicateList[name] = name in this.duplicateList ? this.duplicateList[name] + 1 : 1;
+    if (this.duplicateList[name] > 1) {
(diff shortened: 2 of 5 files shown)


example/missing-n-duplicate.scene
```diff
@@ -0,0 +1,37 @@
...
+      "parent": "New_Node6",
+      "index": 0
+    },
+    "New_Node8": {
+      "parent": "New_Node6",
+      "index": 0
+    }
+  },
+  "inherits": "./ArchitectureKit/Pillar1x3Rnd.gltf"
+}
```
brianpeiris: This doesn't actually demonstrate duplicate nodes.

src/client/editor/ConflictResolution.js
```diff
@@ -0,0 +1,97 @@
+export default class ConflictHandler {
```
brianpeiris: Generally, if a module exports a single default class, I think we want the file name to match the class name.

src/client/editor/ConflictResolution.js
```diff
@@ -0,0 +1,97 @@
...
+
+    if (node.children) {
+      node.children.forEach((child, i) => {
+        child.userData._path = node.userData._path.slice(0);
+        this.findDuplicates(child, layer + 1, i);
+      });
+    }
+  };
+
+  setDuplicateStatus = newStatus => {
```
brianpeiris: If a class member is only used internally, or meant for private use, we start them with an underscore. I think that applies to a few of the methods in this class. Also applies to properties like `duplicateList`.

src/client/editor/ConflictResolution.js
```diff
@@ -0,0 +1,97 @@
...
+  };
+
+  getDuplicateByName = name => {
+    if (!(name in this.duplicateList)) {
+      return false;
+    }
+    return this.duplicateList[name] > 1;
+  };
+
+  getMissingStatus = () => {
```
brianpeiris: Looks like this method is not used anywhere.

src/client/editor/ConflictResolution.js
```diff
@@ -0,0 +1,97 @@
...
+      if (child === scene) {
+        return;
+      }
+      if (!child.userData._isMissingRoot && !child.userData._isDuplicateRoot) {
+        child.userData[propertyName] = child.parent ? child.parent.userData[propertyName] : false;
+      }
+    });
+  };
+
+  checkResolvedMissinRoot = scene => {
```
brianpeiris: typo

src/client/editor/ConflictResolution.js
```diff
@@ -0,0 +1,97 @@
...
+          child.userData._isMissingRoot = false;
+          child.userData._missing = false;
+          resolvedList.push(child);
+        }
+      }
+    });
+    this.setMissingStatus(newStatus);
+    return resolvedList;
+  };
+}
```
brianpeiris: Nice job on refactoring this out into a class! It's much cleaner now.

src/client/ui/panels/HierarchyPanelContainer.js
```diff
@@ -130,15 +130,17 @@ class HierarchyPanelContainer extends Component {
 
   rebuildNodeHierarchy = () => {
     const handler = this.props.editor.scene.userData._conflictHandler;
-    const list = handler.checkResolvedMissinRoot(this.props.editor.scene);
-    if (list.length > 0) {
-      list.forEach(resolvedMissingRoot => {
-        this.props.editor.removeObject(resolvedMissingRoot);
-      });
+    if (handler) {
```
brianpeiris: Hmm, I guess we're doing this because new scenes don't have conflictHandlers? Maybe they should. 
Doesn't have to be fixed in this PR, but this kind of `if` statement (without an `else`) could lead to some surprising bugs. Maybe just keep a mental note to come back to this.

PR #60, opened 2018-07-18
Export using updated THREE.GLTFExporter

robertlong:
Removing editor data, and serializing component data to be added in future PRs.


src/server/index.js
```diff
@@ -183,45 +183,45 @@ export default async function startServer(options) {
...
-    } else if (ctx.request.files && ctx.request.files.file) {
-      const file = ctx.request.files.file;
+      ctx.req.pipe(writeStream);
 
-      await fs.rename(file.path, filePath);
+      await new Promise((resolve, reject) => {
+        function cleanUp() {
+          // eslint-disable-next-line
+          writeStream.removeListener("finish", onFinish);
+          // eslint-disable-next-line
```
brianpeiris: Please use the strict form of this disable comment: `// eslint-disable-next-line no-use-before-define`
Or just configure the rule to allow for use of functions before they're defined.

PR #65, opened 2018-07-22
Enhanced numeric input with bugfixes, scrubbing, scrolling and arrowkeys

### LINKED ISSUE #7: Improvements to numeric property component
- [x] Issues with editing numbers
- [x] "Scrub" to change number
- [ ] Maybe display a slider when there is a fixed range

brianpeiris:
Implemented a `NumericInput` that fixes a lot of the flaws of `react-numeric-input`. This avoids messing with your manual typing as much as possible but still fires `onChange` if it sees a valid number while you're typing.
It allows you to edit partial numbers, negative numbers and decimals as if it were a regular old `<input />` and only attempts to validate and reformat the number when you blur the input or a new prop value comes in. 
It also avoids state changes if the incoming prop value is equivalent to, but not exactly, the current input state, so that your typing isn't interrupted by an `onChange` cycle.

In addition to the fixes, it also provides these features:
- Mouse scrubbing with pointer-lock on the middle mouse button
- Scroll wheel scrubbing
- Arrow key increment/decrement
- Varying step sizes toggled on `Ctrl` and `Shift` keys with any of the above input modes

Closes #7

### DIFF
--- a/src/client/ui/inputs/NumericInput.js
+++ b/src/client/ui/inputs/NumericInput.js
@@ -1,7 +1,165 @@
 import React from "react";
-import styles from "./NumericInput.scss";
-import ReactNumericInput from "react-numeric-input";
+import PropTypes from "prop-types";
 
-export default function NumericInput(props) {
-  return <ReactNumericInput style={false} className={styles.numericInput} {...props} />;
+import styles from "./NumericInput.scss";
+
+function round(value) {
+  return Math.round(value * 1000) / 1000;
+}
+
+const partialValue = /[-.0]$/;
+const wholeNumber = /-?[0-9]+$/;
+
+export default class NumericInput extends React.Component {
+  static propTypes = {
+    value: PropTypes.number,
+    mediumStep: PropTypes.number,
+    smallStep: PropTypes.number,
+    bigStep: PropTypes.number,
+    onChange: PropTypes.func
+  };
+
+  static defaultProps = {
+    smallStep: 0.1,
+    mediumStep: 1,
+    bigStep: 10
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: props.value.toString(),
+      step: props.mediumStep
+    };
+    this.lastValidValue = this.props.value;
+    window.addEventListener("printstate", () => {
+      console.log("react state", this.props, this.state);
+    });
+  }
+
+  getStepForKeys(ctrlKey, shiftKey) {
+    let step = this.props.mediumStep;
+    if (ctrlKey) {
+      step = this.props.smallStep;
+    } else if (shiftKey) {
+      step = this.props.bigStep;
+    }
+    return step;
+  }
+
+  onKeyDown = e => {
+    const { ctrlKey, shiftKey, key } = e;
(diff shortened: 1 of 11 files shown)


src/client/ui/inputs/NumericInput.js
```diff
@@ -1,7 +1,162 @@
...
+    this.state = {
+      value: props.value.toString(),
+      step: props.mediumStep
+    };
+    this.lastValidValue = this.props.value;
+  }
+
+  getStepForKeys(ctrlKey, shiftKey) {
+    let step = this.props.mediumStep;
+    if (ctrlKey) {
```
robertlong: ctrl + click on macOS is equivalent to a right click event. I think we may want to use the alt key instead. I think most programs do the smaller step with alt on mac. Is that also true for Windows?

robertlong: It looks like `ReactNumericInput` used `e.ctrlKey || e.metaKey` https://github.com/vlad-ignatov/react-numeric-input/blob/master/src/NumericInput.jsx#L707

I still kinda think alt makes more sense though. I use command to jump around inside inputs.

src/client/ui/inputs/NumericInput.js
```diff
@@ -1,7 +1,162 @@
...
-  return <ReactNumericInput style={false} className={styles.numericInput} {...props} />;
+function round(value) {
+  return Math.round(value * 1000) / 1000;
+}
+
+const partialValue = /[-.0]$/;
+const wholeNumber = /-?[0-9]+$/;
+
+export default class NumericInput extends React.Component {
+  static propTypes = {
```
robertlong: ReactNumeric input also had support for min/max, precision, and format. I think we'll want to add those options eventually. We may also want to add them as options to the component schema?

src/client/ui/inputs/NumericInput.js
brianpeiris: Yeah, I tried `alt` but as far as I can tell, it always triggers the menu bar in Firefox and there's no way to prevent it. I'll try again, but if that doesn't work, maybe I can change it to metaKey for macOS alone.

src/client/ui/inputs/NumericInput.js
brianpeiris: Yup, adding min/max right now -- we definitely need it for things like the metallic and roughness values. Not sure precision needs to be an option right now. Precision should probably be the same for all the numbers we care about in the editor, should probably never be more than 3 digits.
I can't think of a usecase for format right now, but we can add it in the future if necessary.

src/client/ui/inputs/NumericInput.js
robertlong: If you're going to go for something that check what platform you're on`alt`would be preferrable to `cmd` on mac. However, `e.ctrlKey || e.metaKey` would probably be just fine.

src/client/ui/inputs/NumericInput.js
robertlong: The only thing I was thinking is maybe for adding the degrees symbol when the input isn't focused?

PR #76, opened 2018-07-24
Make URIs and URLs consistent

### LINKED ISSUE #64: File paths are inconsistent when round-tripping scenes
Things that refer to file paths (references, materials, material textures) can have very inconsistent file paths at the moment. I've seen everything from `./foo.file` to `/api/files/foo.file` to `http://hubs.local:9090/api/files/foo.file` at various stages of adding, setting and saving components/scenes.
It can sometimes lead to scenes breaking.

brianpeiris:
- Ensure URIs are only relative on disk.
- In-memory state always uses absolute URLs
- Absolute URLs are displayed as relative URLs in FileInput and SaveableComponent PropertyGroups.
- Clone scene before serializing to avoid leaking relative URLs into in-memory state.
- Avoid reloading scene on save with a ignoreNextSceneFileChange
- Fix issues with scene reload
- Some refactoring/cleanup in Editor.js

Closes #64


src/client/editor/components/TransformComponent.js
```diff
@@ -21,12 +21,15 @@ export default class TransformComponent extends BaseComponent {
     };
   }
 
-  get props() {
-    return TransformComponent._propsFromObject(this._object);
+  constructor(node, object) {
+    super(node, object);
+    Object.defineProperty(this, "props", {
```
robertlong: What is this for?

src/client/api/project.js
```diff
@@ -1,6 +1,16 @@
 import EventEmitter from "eventemitter3";
 
-export default class Project extends EventEmitter {
+const projectDirectoryPath = "/api/files/";
+
+export function getRelativeURI(uri) {
```
robertlong: Because we are eventually going to want to make another project adapter for a hosted version of the editor, I'd rather these be instance methods on the Project class and pass the project class to the necessary components via the context API.

src/client/editor/components/TransformComponent.js
brianpeiris: TransformComponent is special in that it doesn't store its own position, rotation and scale properties, but it reads it directly from the Object3D

robertlong: Maybe leave a comment there. Seems good though.

PR #77, opened 2018-07-25
Combine meshes and remove empty nodes

robertlong:
- Adds the static property to the node.
- Combines meshes that are marked static and share the same material and shadow properties
- Adds the Mesh component (Helps to identify meshes in the editor and could be used for other things in the future)
- Removes empty nodes that are marked as static after combining meshes.


src/client/editor/SceneLoader.js
```diff
@@ -176,8 +303,15 @@ function inflateGLTFComponents(scene, addComponent) {
       }
     }
 
-    if (object instanceof THREE.Mesh && object.material instanceof THREE.MeshStandardMaterial) {
-      addComponent(object, "standard-material", undefined, true);
+    if (object instanceof THREE.Mesh) {
+      addComponent(object, "mesh", undefined, true);
```
brianpeiris: We should avoid passing or assigning `undefined`, it gets confusing when you care about whether a method was intentionally called with no value or whether it was never passed in in the first place. Use `null`.

Same for the two `addComponent`s below.

src/client/editor/SceneLoader.js
```diff
@@ -64,41 +66,185 @@ function loadGLTF(url) {
...
-  if (node.extras) {
-    for (const key in node.extras) {
-      console.log(key);
+  const componentsToExport = Components.filter(c => !c.dontExportProps).map(component => component.componentName);
+
+  // Second pass at scene optimization.
+  clonedScene.traverse(object => {
+    const userData = object.userData;
+
+    // Move component data to userData.components
```
brianpeiris: Can we call it `spokeComponents` now?

src/client/editor/components/ShadowComponent.js
```diff
@@ -4,18 +4,28 @@ import { types } from "./utils";
...
-      if (obj instanceof THREE.Mesh) {
-        obj[propertyName] = value;
-        obj.material.needsUpdate = true;
-      }
-    });
+    this._object[propertyName] = value;
+
+    if (this._object.material) {
+      this._object.material.needsUpdate = true;
+    }
```
brianpeiris: Hmm, I guess we should actually disallow putting this component on a non-Mesh object. 
Probably a task for after alpha.

src/client/ui/inputs/BooleanInput.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+export default function BooleanInput({ onChange, value }) {
+  return <input type="checkbox" checked={value} onChange={e => onChange(e.target.checked)} />;
+}
+
+BooleanInput.defaultProps = {
+  value: "",
```
brianpeiris: Should be `false`?

src/client/editor/SceneLoader.js
```diff
@@ -64,41 +66,185 @@ function loadGLTF(url) {
...
+
+      clonedScene.add(combinedMesh);
 
-        node.extras.components[component.name] = component.props;
+      for (const mesh of meshes) {
+        const meshIndex = mesh.parent.children.indexOf(mesh);
+        const parent = mesh.parent;
+        mesh.parent.remove(mesh);
+        const replacementObj = new THREE.Object3D();
+        replacementObj.copy(mesh, true);
```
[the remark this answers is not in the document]
robertlong: Oh, yeah that seems much better!

src/client/editor/SceneLoader.js
```diff
@@ -64,41 +66,185 @@ function loadGLTF(url) {
...
+    // Remove empty components object.
+    if (object.userData.components && isEmptyObject(object.userData.components)) {
+      delete object.userData.components;
+    }
+  });
+
+  removeObjects(
+    clonedScene,
+    (object, canBeRemoved) =>
+      canBeRemoved &&
```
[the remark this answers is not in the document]
robertlong: Yeah, agreed. Although if you ever want to traverse a scene and remove objects based on some condition this is the method I'd suggest.

src/client/editor/SceneLoader.js
robertlong: Probably not for the `userData` property. We want the components to line up with how Hubs expects them for now. We'll then probably want to move component data to individual extensions still or at minimum, namespaced extras fields.

PR #81, opened 2018-07-25
Warning display - missing resources

tony056:
- Added cases:
  - missing inherited file (.scene/.gltf)
  - missing scene reference file (.scene/.gltf)
  - missing material file (.material)

- Warning display types:
  - FileInput warning (scene reference, .material)
  - SystemMessage warning (inherited file)


src/client/editor/SceneLoader.js
```diff
@@ -538,6 +541,9 @@ export async function loadScene(uri, addComponent, isRoot = true, ancestors) {
   }
 
   const sceneResponse = await fetch(url);
+  if (!sceneResponse.ok) {
+    throw url;
```
brianpeiris: It's generally bad practice to `throw` an object that is not an `Error`. Best to use a custom error object like `SceneLoadingError`, as I mentioned above.

src/client/editor/components/BaseComponent.js
```diff
@@ -21,6 +21,13 @@ export default class BaseComponent {
     this.props[propertyName] = value;
   }
 
+  updateResourceValidation(target, value) {
+    const resourcesValidation = !this._object.resourcesValidation ? {} : this._object.resourcesValidation;
+    resourcesValidation[target] = value;
+    //this.updateProperty("_resourcesValidation", resourcesValidation);
+    this._object.resourcesValidation = resourcesValidation;
+  }
```
brianpeiris: I don't think `BaseComponent` should have a method called `updateResourceValidation`. Most components will not have resources associated with them. It also does not make sense to store `resourcesValidation` on `_object` -- it is associated with the component, specifically the component's property, not with the `_object`.

src/client/editor/components/StandardMaterialComponent.js
```diff
@@ -62,6 +62,8 @@ export default class StandardMaterialComponent extends SaveableComponent {
       this._object[map] = texture;
     } catch (e) {
       // TODO Should show warning on texture property when this happens.
+      const resource = this._mapTypesToTextureType(map);
```
brianpeiris: I would just pass the original `propertyName` in from `updateProperty`, instead of doing a reverse lookup again.

src/client/ui/EditorContainer.js
```diff
@@ -368,9 +374,38 @@ class EditorContainer extends Component {
...
-      console.error(e);
-    });
+    this.props.editor
+      .openRootScene(uri)
+      .then(scene => {
+        document.title = `Hubs Editor - ${scene.name}`;
+      })
+      .catch(e => {
+        this.setState({
+          systemMessage: {
```
brianpeiris: This should use Robert's new `DialogContext` with the `openModal` state property. In fact, I think you can just use the `OptionDialog` and ditch `SystemMessageModalContainer`

src/client/ui/PropertyGroup.js
```diff
@@ -25,9 +27,17 @@ function PropertyGroup(props) {
 
   const renderSaveButtons = () => {
     if (!saveable) return;
+    const srcStatus = {
+      class: srcIsValid ? styles.src : styles.invalidSrc,
+      tip: srcIsValid ? "file path" : "Cannot find the file.",
```
brianpeiris: I don't think the "file path" tooltip is necessary when there is no issue with the `src`.

brianpeiris: Same goes for "file path" in `FileInput` below.

src/client/ui/inputs/FileInput.scss
```diff
@@ -14,3 +14,20 @@
   background: none;
   padding: 0;
 }
+
+:local(.invalidPath) {
+  border: 2px solid $red;
+  border-radius: 2px;
```
brianpeiris: Get rid of the border-radius, since the original input doesn't have one.

src/client/ui/inputs/FileInput.scss
```diff
@@ -14,3 +14,20 @@
   background: none;
   padding: 0;
 }
+
+:local(.invalidPath) {
+  border: 2px solid $red;
```
brianpeiris: Use `outline` so that the layout isn't shifted when there's an error.

src/client/ui/modals/SystemMessageModalContainer.js
```diff
@@ -0,0 +1,81 @@
...
+  );
+}
+
+SystemMessageModalContainer.propTypes = {
+  messageType: PropTypes.string,
+  messageContent: PropTypes.any,
+  isOpen: PropTypes.bool,
+  onRequestClose: PropTypes.func,
+  actions: PropTypes.array
+};
```
brianpeiris: As mentioned above, all of this can be replace by `OptionDialog` now.

src/client/editor/components/StandardMaterialComponent.js
```diff
@@ -62,6 +62,8 @@ export default class StandardMaterialComponent extends SaveableComponent {
...
+      this.updateResourceValidation(resource, false);
```
brianpeiris: Actually, this was incorrect to begin with. This `try-catch` doesn't do anything. Validation handling should be in the `onError` callback above.

src/client/ui/panels/PropertiesPanelContainer.js
```diff
@@ -158,12 +159,21 @@ class PropertiesPanelContainer extends Component {
...
     switch (prop.type) {
       case types.number:
         return { min: prop.min, max: prop.max };
-      case types.file:
-        return { openFileDialog: this.props.openFileDialog, filters: prop.filters };
+      case types.file: {
+        const isValid =
+          component._object && component._object.resourcesValidation
+            ? component._object.resourcesValidation[prop.name]
+            : true;
```
brianpeiris: This feels wrong. You're mixing meanings here. The extras are supposed to be based off the prop schema, not the component instance state itself, and it doesn't seem right to store validation on the object anyway, as I mentioned above.

tony056:
- Used `{path: "./abc.gltf", isValid: true}` for `component.src` and `component.props.src`
- Used  outline
- Handled Error with the error object.

brianpeiris:
Looks good. Thanks for all the changes. Spotted a couple of bugs that I'll point out in person

PR #92, opened 2018-07-27
Added error and progress dialogs. Prevent multiple clicks of export/save

### LINKED ISSUE #86: File dialogs error when save / export are clicked multiple times
The save/export button should be disabled or some sort of progress bar should be shown when saving/exporting.

robertlong:
Fixes #86, present users with errors on saving/loading/exporting, show progress dialog on save/load/export.

Todo:
- [x] Style dialog

In the future, the progress dialog can instead be a passive spinner that shows the file is being saved in the background.


src/client/ui/EditorContainer.js
```diff
@@ -70,7 +72,6 @@ class EditorContainer extends Component {
         },
         properties: {
           component: PropertiesPanelContainer,
-          props: { openFileDialog: this.openFileDialog },
```
brianpeiris: The material save/load buttons are broken now. You removed `openFileDialog` here but you didn't make the corresponding changes in `PropertiesPanelContainer`.

src/client/ui/EditorContainer.js
```diff
@@ -261,6 +260,11 @@ class EditorContainer extends Component {
   };
 
   serializeAndSaveScene = async sceneURI => {
+    this.showDialog(ProgressDialog, {
+      title: "Saving Scene",
+      message: "Saving scene..."
+    });
```
brianpeiris: This flashes the dialog briefly, since the save operation is usually quick after the first save.

src/client/ui/EditorContainer.js
```diff
@@ -285,61 +289,73 @@ class EditorContainer extends Component {
...
+        title: "Error Exporting Scene",
+        message: e.message
+      });
+
+      console.error(e);
+    }
 
-    // Close modal
-    this.setState({ openModal: null });
+    this.hideDialog();
```
brianpeiris: This hides the error dialog immediately in the error case.

src/client/ui/EditorContainer.js
```diff
@@ -86,6 +87,8 @@ class EditorContainer extends Component {
           }
         }
       },
+      DialogComponent: null,
+      dialogProps: {},
       openModal: null,
```
brianpeiris: I think you can kill all the `openModal` stuff now.

src/client/ui/EditorContainer.js
```diff
@@ -285,61 +289,73 @@ class EditorContainer extends Component {
...
     } catch (e) {
-      throw e;
+      this.showDialog(ErrorDialog, {
+        title: "Error Saving Scene",
+        message: e.message
+      });
+      console.error(e);
     }
+
+    this.hideDialog();
```
brianpeiris: This hides the error dialog immediately in the error case.

src/client/ui/dialogs/FileDialog.scss
```diff
@@ -1,9 +1,3 @@
-:local(.fileDialogModalContainer) {
-  display: flex;
-  flex-direction: column;
-  flex: 1;
-}
```
brianpeiris: FileDialog should probably set a min-width and min-height styles. It gets really small at times, like when you navigate to an empty folder.

src/client/ui/EditorContainer.js
brianpeiris: This also broke the file dialog in FileInput

robertlong: Ah, right I forgot about that. Thanks!

src/client/ui/EditorContainer.js
robertlong: Yeah, what's the best way to handle this? Should we remove it entirely and just hide the dialog as soon as you confirm? Maybe we need to keep track of whether or not you are currently saving and cancel the current save action?

src/client/ui/EditorContainer.js
robertlong: I think we're going to use that modal system for the export modal. We can move to using a context based API though.

src/client/ui/EditorContainer.js
brianpeiris: I don't think it needs to be complicated. Usually you'd do something like:
```javascript
let saved = false;
setTimeout(() => {
  if (saved) return;
  this.showDialog(ProgressDialog, {
    title: "Saving Scene",
    message: "Saving scene..."
  });
}, 100);
```
and then set `saved = true` at the bottom of the `try`

I don't know if we need to do more to communicate that it _was_ successfully saved, any more than just the `*` disappearing from the title bar.

src/client/ui/panels/PropertiesPanelContainer.js
```diff
@@ -119,51 +121,96 @@ class PropertiesPanelContainer extends Component {
...
+
+          component.src = src;
+          component.shouldSave = true;
+          component.modified = false;
+          component.constructor.inflate(this.state.object, await this.props.project.readJSON(component.src));
+          this.props.editor.signals.objectChanged.dispatch(this.state.object);
+          this.hideDialog();
+        } catch (e) {
+          console.error(e);
+          this.showDialog(ErrorDialog, {
```
brianpeiris: This is bugged. All of the above `this.hideDialog()` calls and `this.showDialog()` calls should be `this.props.hide/showDialog()`

Issue #88, opened 2018-07-26
ColorInput doesnt respect PropertiesPanel scroll position


robertlong:
![screen shot 2018-07-25 at 5 34 26 pm](https://user-images.githubusercontent.com/1753624/43234481-18ab1b2a-9031-11e8-90cf-f189de1bf933.png)

tony056:
PR #106

PR #106, opened 2018-07-30
Color Picker Position

tony056:
- made the color picker respect to the container.
- made the click cover not to overlap with the scrollbar of the property panel.


tony056:
resolved #88

PR #121, opened 2018-07-31
Make static mode use inheritance and serialize it properly

robertlong:
This PR moves static mode from a boolean field to an enum.

There are now 3 static modes:
- Static: This object doesn't change and is free to combined if the proper conditions are met.
- Dynamic: This object may change or shouldn't be combined.
- Inherits: This object should use the static mode of its parent. (Default)


src/client/editor/SceneLoader.js
```diff
@@ -73,12 +81,25 @@ function shallowEquals(objA, objB) {
   return true;
 }
 
-function isEmptyObject(obj) {
-  for (const key in obj) {
-    return false;
+function hasExtrasOrExtensions(obj) {
+  for (const key in obj.userData) {
+    if (!key.startsWith("_")) {
```
brianpeiris: Check for has `hasOwnProperty` here, otherwise this will return true for all objects because of keys like `'__proto__'`.

src/client/editor/SceneLoader.js
```diff
@@ -73,12 +81,25 @@ function shallowEquals(objA, objB) {
...
 
-  return true;
+  return false;
+}
+
+function removeEditorData(scene) {
+  scene.traverse(({ userData }) => {
+    // Remove editor data.
+    for (const key in userData) {
+      if (key.startsWith("_")) {
```
brianpeiris: Check for `hasOwnProperty`.

PR #125, opened 2018-08-01
Copy default project files into new project.

### LINKED ISSUE #120: CLI should create directory path if it doesn't exist and initialize it with the example files
You should be able to run `spoke ~/a/path/of/your/choosing/parthenon/`

robertlong:
Fixes #120


bin/spoke
```diff
@@ -23,22 +26,43 @@ if (process.env.NODE_ENV === "development") {
...
 
-startServer({
-  projectPath,
-  port: commander.port,
-  https: commander.https
-})
-  .then(() => {
-    console.log(`Server listening on port ${commander.port}`);
+  rl.question("Creating new Spoke project. Would you like to add a the default assets? (y/n)", answer => {
+    const copyDefaultAssets = answer.toLowerCase() === "y";
```
brianpeiris: Maybe `.trim()` also?

brianpeiris: Can we have this be "y" by default, so that the user just has to hit enter?

bin/spoke
```diff
@@ -23,22 +26,43 @@ if (process.env.NODE_ENV === "development") {
...
+    copyDefaultAssets,
+    port: commander.port,
+    https: commander.https
+  })
+    .then(() => {
+      console.log(`Server listening on port ${commander.port}`);
+
+      if (commander.open) {
+        opn(`http://localhost:${commander.port}`);
+      }
```
brianpeiris: Maybe open by default, or ask here as well?

robertlong: Open by default seems like the correct thing to do.

PR #133, opened 2018-08-01
AssetExplorer - Extend .gltf/.scene

tony056:
- Add:
  - `extend` button in the contextmenu of AssetExplorerPanel.


src/client/editor/Editor.js
```diff
@@ -378,6 +378,26 @@ export default class Editor {
...
   }
 
+  async extendScene(inheritedURI) {
+    const extendSceneDef = {
+      entities: {},
+      inherits: inheritedURI
+    };
+
+    this.deselect();
+    this._resetHelpers();
```
brianpeiris: You need `_deleteSceneDependencies()` and `_clearCaches()` in here as well. You're basically creating a new scene, so you can follow `loadNewScene` as an example -- I would try to maintain the same order of operations.

src/client/editor/Editor.js
```diff
@@ -378,6 +378,26 @@ export default class Editor {
...
+    const ancestors = [];
+    const scene = await loadSerializedScene(extendSceneDef, inheritedURI, this.addComponent, true, ancestors);
+
+    scene.userData._ancestors = ancestors;
```
brianpeiris: No need to do this, it happens in SceneLoader already.

src/client/editor/Editor.js
```diff
@@ -378,6 +378,26 @@ export default class Editor {
...
+
+    this._setSceneInfo(scene, null);
+    this.scenes = [this.sceneInfo];
+    this._setSceneDefaults(scene);
+    this._setScene(scene);
+    this._addDependency(null, scene);
```
brianpeiris: It doesn't make sense to add a dependency on a null uri, it should be inheritedURI instead of null.

src/client/editor/Editor.js
```diff
@@ -378,6 +378,26 @@ export default class Editor {
...
+    return scene;
```
brianpeiris: Dispatch `sceneModified` before returning.

src/client/ui/panels/AssetExplorerPanelContainer.js
```diff
@@ -188,6 +190,18 @@ class AssetExplorerPanelContainer extends Component {
...
   };
 
+  onExtend = (e, file) => {
+    if (!file) {
+      return;
+    }
+    if (file.ext === ".gltf") {
+      this.props.editor.signals.openScene.dispatch(file.uri);
+    } else {
+      this.props.editor.extendScene(file.uri);
```
brianpeiris: You can't call `extendScene` here directly, it has to go through a signal dispatch similar to `openScene` which is handled in EditorContainer so that you can call `confirmSceneChange`, or else you will blow away the user's unsaved changes without notifying them. I would create a new `extendScene` signal for this.

src/client/editor/Editor.js
tony056: It seems that it does only in `loadScene`, but not `loadSerializedScene`, cuz when I use `loadSerializedScene` without `ancestors` here, the hierarchy panel doesn't display any ancestor files on the top.

brianpeiris: `loadSerializedScene` calls `loadScene`. You do have to pass an empty array in, but you don't have to set `userData._ancestors` yourself.

src/client/ui/EditorContainer.js
```diff
@@ -428,7 +429,17 @@ class EditorContainer extends Component {
...
     if (this.props.editor.sceneInfo.uri === uri) return;
     if (!this.confirmSceneChange()) return;
+    const action = "openRootScene";
+    this._tryLoadSceneFromURI(uri, action, this.onOpenScene);
+  };
 
+  onExtendScene = async uri => {
+    if (!this.confirmSceneChange()) return;
+    const action = "extendScene";
+    this._tryLoadSceneFromURI(uri, action, this.onExtendScene);
```
brianpeiris: Would have been better to just pass in the function itself, instead of just the function name. Generally good to avoid dynamic, string-based lookups like this where possible. 
i.e:
```javascript
this._tryLoadSceneFromURI(uri, this.props.editor.extendScene.bind(this.props.editor), this.onExtendScene)
```

or just 

```javascript
this._tryLoadSceneFromURI(uri, this.props.editor.extendScene, this.onExtendScene)
```
if you convert `extendScene` to and arrow function first.

PR #145, opened 2018-08-03
Naming Conflicts

tony056:
- Cases:
  - add/duplicate node -> automatically add postfix to resolve naming conflicts.
  - renaming node -> pop out the error dialog and reset the node name back.
  - importing .gltf -> pop out the error dialog when there are any duplicate/empty node names
    - overwrite the .gltf when the user press 'okay'
  - case 3 doesn't include the missing node names situation since the system does not know where it should put orphan nodes.
  - auto resolve conflicts while loading scene ref.


src/client/ui/EditorContainer.js
```diff
@@ -451,35 +437,91 @@ class EditorContainer extends Component {
...
+          }
+        });
+      }
+    } finally {
+      opened = true;
+    }
+  };
+
+  onSceneErrorOccurred = (error, uri, reload) => {
+    if (error.type === "import") {
```
brianpeiris: Can probably use `instanceof` here.

src/client/ui/EditorContainer.js
```diff
@@ -451,35 +437,91 @@ class EditorContainer extends Component {
...
       });
-    } finally {
-      opened = true;
+    } else {
+      // renaming
+      this.showDialog(ErrorDialog, {
+        title: "Name in Use",
+        message: "Node name is already in use. Please choose another.",
+        confirmLabel: "Okay"
+      });
```
brianpeiris: When does this happen? Maybe turn the `else` into an `else if` to make it explicit?

src/client/ui/EditorContainer.js
```diff
@@ -451,35 +437,91 @@ class EditorContainer extends Component {
...
-              ${e.url}.
-              Please make sure the file exists and then press "Resolved" to reload the scene.
-            `,
+            title: "Resolve Node Conflicts",
+            message:
+              "We've found duplicate and/or missing node names in this file.\nWould you like to fix all conflicts?\n*This will modify the original file.",
             options: [
               {
-                label: "Resolved",
+                label: "okay",
```
brianpeiris: Use consistent casing, "Okay" and "Cancel" below.

src/client/ui/EditorContainer.js
```diff
@@ -451,35 +437,91 @@ class EditorContainer extends Component {
...
+        title: "Name in Use",
+        message: "Node name is already in use. Please choose another.",
+        confirmLabel: "Okay"
+      });
+    }
+  };
+
+  _overwriteConflictsInSource = async (uri, conflictHandler, callback) => {
+    const { project } = this.props;
+    if (uri && uri.endsWith(".gltf")) {
```
brianpeiris: Should we do something in an `else` condition?

brianpeiris:
Found a bug:
1. Create a new scene
2. Add two nodes such that the scene contains "New_Node" and "New_Node_1"
3. Save scene
4. Refresh and load the scene
5. Create a new node
6. New node has a duplicate name

src/client/editor/Editor.js
```diff
@@ -266,6 +274,10 @@ export default class Editor {
 
   _setScene(scene) {
     this.scene = scene;
+    this._conflictHandler = new ConflictHandler();
+    if (this.userData && this.userData._conflictHandler) {
```
brianpeiris: Should be `this.scene.userData`

brianpeiris:
Found another bug:
1. Load Spoke
1. Click File -> New Scene
1. Drop in a gltf model
1. Click "Edit Prefab" on model
1. Return to unsaved scene
1. "TypeError: this._conflictHandler is undefined" in console.

src/client/editor/Editor.js
```diff
@@ -180,7 +188,7 @@ export default class Editor {
       if (previousURI.endsWith(".gltf")) {
         const name = last(poppedURI.split("/"));
         this._prefabBeingEdited.name = name;
-        this._duplicateNameCounters.set(name, this._duplicateNameCounters.get(name) || 0);
+        this._conflictHandler.addToDuplicateNameCounters(name);
```
brianpeiris: It should also de-duplicate the name here. Otherwise this leads to a bug:
1. Load Spoke
1. Add a Node. Name it "test.scene"
1. Drop in a GLTF model
1. "Edit Prefab" on the model
1. Save the prefab as "test.scene"
1. Return to the unsaved scene
1. There are now two "test.scene" nodes.

brianpeiris:
Sorry, found another bug:
1. Load Spoke
1. Add a new node
1. Manually name it "test_1"
1. Add a new node
1. Try to name it "test"
1. It won't let you.

brianpeiris:
I think this bug is the same root cause as the previous comment, but I'll write out the steps for reference:
1. Load Spoke
1. Add three new nodes
1. Change the name on the last node from "New_Node_2" to "New_Node_1"
1. It doesn't stop you from doing that, so now there are two "New_Node_1" nodes in the scene.

PR #152, opened 2018-08-10
Add slidebar

tony056:
- Added slider
slider

### DIFF
--- a/src/client/ui/inputs/CompoundNumericInput.js
+++ b/src/client/ui/inputs/CompoundNumericInput.js
@@ -6,7 +6,7 @@ import styles from "./CompoundNumericInput.scss";
 
 export default function CompoundNumericInput({ value, onChange, ...extras }) {
   const { min, max } = extras;
-  const step = 0.1;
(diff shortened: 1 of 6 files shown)


src/client/ui/inputs/CompoundNumericInput.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Slider from "rc-slider";
+import NumericInput from "./NumericInput";
+import styles from "./CompoundNumericInput.scss";
+
+export default function CompoundNumericInput({ value, onChange, ...extras }) {
+  const { min, max } = extras;
+  const step = max && min ? (max - min) / 10 : 0.1;
```
robertlong: We can probably add step to the component definitions and pass step in through the extras property, the default step can be 0.1.

tony056:
- Made step configurable.
- Noticed that there are many `number` in component definitions without `min` and `max`, which can have some troubles while using the slider to adjust the value. I'm not sure the value range of those properties, so I won't do those changes in this PR.

Issue #198, opened 2018-08-30
Add the ability to create box colliders in the editor


brianpeiris:
A user should be able to import a whole model into Spoke (e.g. a model of a room from Sketchfab) and then add various
box colliders to the scene. 

Spoke should show a helper gizmo for the collider that lets you scale and position it.

PR #200, opened 2018-08-30
Feature/box collider

brianpeiris:
Add a box collider that you can transform with the existing transform controls and exports as a component with transform-like properties (position, scale, rotation).


brianpeiris:
Closed #198

PR #207, opened 2018-09-09
Import models from the web

### LINKED ISSUE #206: Resolve, download and import models from the web
If users enter a URL, its assumed to be a GLB/GLTF URL or a Sketchfab/Poly URL which needs to be resolved via reticulum. For the URL case, the asset is downloaded into a new "imports" folder in the project folder and linked in automatically.

brianpeiris:
- Uses reticulum to resolve model URLs dropped onto the viewport
- Imports model into project folder using a CRC32 hash of the origin url.
- Unzips sketchfab models if necessary

Closes #206


src/client/ui/panels/ViewportPanelContainer.js
```diff
@@ -40,15 +41,35 @@ class ViewportPanelContainer extends Component {
...
+      );
+      if (!urlItem) return;
+      try {
+        this.props.showDialog(ProgressDialog, {
+          title: "Importing Asset",
+          message: "Importing asset..."
+        });
+        const url = (await new Promise(resolve => urlItem.getAsString(resolve))).trim();
+        const importedUri = await this.props.editor.project.import(url);
+        this.props.editor.addGLTFModelNode("Model", importedUri);
```
gfodor: Might be nice to try to name this better -- is there any metadata we can pull from the GLTF file?

gfodor: Or, if not, give it a unique name like "Model #xxxx" or something?

src/server/index.js
```diff
@@ -295,6 +328,48 @@ export default async function startServer(options) {
...
+      const { raw, meta } = await fetch(`https://${mediaServer}/api/v1/media`, {
+        method: "POST",
+        headers: { "content-type": "application/json" },
+        body: JSON.stringify({ media: { url: origin, index: 0 } })
+      }).then(r => r.json());
+
+      await fs.ensureDir(path.join(projectPath, "imported"));
+
+      const resp = await fetch(raw);
+      if (meta && meta.expected_content_type.includes("gltf+zip")) {
```
gfodor: `expected_content_type` may be undefined

src/client/ui/panels/ViewportPanelContainer.js
brianpeiris: Ah, I didn't think to look in the GLTF. Looks like sketchfab might put the model name in there, and Google Poly gives us a displayName in the API.

PR #213, opened 2018-09-13
Feature/login and publish

brianpeiris:


src/client/ui/dialogs/LoginDialog.js
```diff
@@ -0,0 +1,66 @@
...
+
+  render() {
+    const { onLogin, authStarted, onCancel, hideDialog } = this.props;
+    return (
+      <div className={styles.dialogContainer}>
+        <Header title="Login to Publish" />
+        <div className={styles.loginContainer}>
+          {authStarted ? (
+            <div className={styles.content}>
+              <div className={styles.message}>Email sent! Please click on the link in th email to continue.</div>
```
gfodor: Suggest we add a spinner here maybe?

gfodor: also probably could have a cancel button -- not strictly necessary. if you enter the wrong email address you are stuck basically

src/client/ui/dialogs/PublishDialog.js
```diff
@@ -0,0 +1,59 @@
...
+  render() {
+    const { onPublish, onCancel, hideDialog } = this.props;
+    return (
+      <div className={styles.dialogContainer}>
+        <Header title="Publish to Hubs" />
+        <div className={styles.publishContainer}>
+          <div className={styles.content}>
+            <img className={styles.sceneThumbnail} />
+            <div>
+              <div className={styles.inputField}>
```
gfodor: Additional field I forgot to add to the db schema is "Composed By Display Name"

gfodor: probably optional (can be left blank)

src/client/ui/dialogs/LoginDialog.js
```diff
@@ -0,0 +1,66 @@
...
+            </div>
+          ) : (
+            <div>
+              <div className={styles.content}>
+                <div className={styles.message}>
+                  Login to publish your scene. You will be sent an email with a magic link.
```
gfodor: "Log in" I think

src/client/editor/Editor.js
```diff
@@ -1701,4 +1706,55 @@ export default class Editor {
   redo() {
     this.history.redo();
   }
+
+  async publishScene() {}
```
gfodor: one thing we probably need to figure out (maybe for next PR) is gathering attribution and potentially also origin information for all composed assets, and submitting it to the scene API. also, reminder that we need to parameterize this so it will either create or update if the scene already has a scene ID

src/client/ui/dialogs/LoginDialog.js
brianpeiris: Hmm, not so sure about the spinner. I think it might be misinterpreted as the app actually doing something that you have to wait for.

src/client/ui/dialogs/PublishDialog.js
brianpeiris: Made #220 for it.

PR #214, opened 2018-09-13
Feature/basic ui changes

gfodor:
Adds a new object creation workflow and removes a bunch of UI elements around component management.

### DIFF
--- a/src/client/ui/Button.scss
+++ b/src/client/ui/Button.scss
@@ -14,6 +14,14 @@
   text-align: center;
   cursor: pointer;
 
+  &:disabled {
+    background: $panel2;
+
+    &:hover {
+      background-color: $panel2;
+    }
+  }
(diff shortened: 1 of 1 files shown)


src/client/ui/panels/HierarchyPanelContainer.js
```diff
@@ -195,10 +195,10 @@ class HierarchyPanelContainer extends Component {
   renderHierarchyNodeMenu = props => {
     const node = props.trigger;
     const hasParent = node && node.object.parent;
+    if (!hasParent) return null;
```
brianpeiris: One reason we might not want to hide the Scene node is that we still might want to give users the ability to mark the entire scene as "Static". The default is "Dynamic". But perhaps we can do a reasonable default thing here instead.

PR #236, opened 2018-09-18
Feature/publish styling

gfodor:
- Turns back on mosaic titles + tweaks docs in light of that
- Styling pass on login flow
- Slightly better model loading error messages from Sketchfab + Poly


gfodor:
no need for a deep review, one q i had is if the break out of the UI flow into `import.js` is the way you'd do it.

brianpeiris:
Since AddNodeActionButtons is a child of ViewportPanelContainer, I would just expose `onURLEntered` as a property of AddNodeActionButtons and have a handler in ViewportPanelContainer. That way you don't have to split it out to import.js

brianpeiris:
LGTM otherwise

gfodor:
I renamed it to `wasModified` but then realized the logic was off for the load file case. I refactored in (and landed it on master) to be `isDefaultScene`, and its `true` iff the scene is an unmodified default scene. (Which is what I wanted anyway)

brianpeiris:
:+1: much clearer now

PR #240, opened 2018-09-19
Feature/scene publish ux

gfodor:
This PR adds a scene "metadata" concept which is now exclusively used to store the publish information for a scene (including the scene id that was last used.) This metadata is stored in the spoke file. It's also incidentally wiped out if you do a "Save As", so the publish information doesn't carry over if you make an in-Spoke copy of a scene. (This is more of a UX thing, we can remove it if you think it's too confusing, but it feels right.)

This also styles the publish flow and makes it so the scene name is effectively the "key" the user is exposed to wrt create vs update. (So, there's no way via the UI to rename a scene.) If the user wants to roll over to a new scene in Hubs, they can either do a Save As (making a new spoke file, which resets the metadata) or there is a "New Scene" button in the publish flow that lets them fill out the full form. 

Other small UX things: the publish button is hidden until you edit the default scene, and you can't publish dirty projects (you are prompted to save before publishing and must save to continue.) This PR also cleans up the ordering and styling of the modal buttons.


src/client/editor/Editor.js
```diff
@@ -123,12 +124,12 @@ export default class Editor {
     this._ignoreSceneModification = false;
     this.signals.sceneGraphChanged.add(() => {
       if (this._ignoreSceneModification) return;
-      this.sceneInfo.modified = true;
+      this.sceneInfo.modified = this.sceneInfo.saved = true;
```
brianpeiris: Hmm, this is confusing. The scene is considered "saved" when it is modified? This can't be right.

src/client/editor/Editor.js
```diff
@@ -109,7 +109,8 @@ export default class Editor {
       scene: new THREE.Scene(),
       helperScene: new THREE.Scene(),
       helpers: {},
-      modified: false
+      modified: false,
+      saved: false
```
brianpeiris: Currently, we determine whether a scene has been saved by checking the existence of `sceneInfo.uri`. 
For example: https://github.com/MozillaReality/Spoke/blob/master/src/client/ui/EditorContainer.js#L296

I would suggest either using that method or using this new flag everywhere.

brianpeiris:
Oh. I get it now. I would rename `sceneInfo.saved` to `sceneInfo.wasModified`.

PR #245, opened 2018-09-20
Feature/login flow info

gfodor:
- Updates the publish flow to tell the user about ToS/PP
- Adds option for users to allow remixing and/or promotion of their scenes
- Uploads the scene file as part of publishing
- Adds the origin URL for imported assets to the GLTF component data in the scene file
- Adds publish command to the File menu in addition to the big button

Relies upon https://github.com/mozilla/reticulum/pull/66

### DIFF
--- a/src/client/editor/Editor.js
+++ b/src/client/editor/Editor.js
@@ -1190,14 +1190,14 @@ export default class Editor {
     this.select(object);
   }
 
-  async addGLTFModelNode(name, uri) {
+  async addGLTFModelNode(name, uri, originUri) {
     const attribution = await this.project.getImportAttribution(uri);
-    this.addUnicomponentNode(name, "gltf-model", true, { src: uri, attribution });
+    this.addUnicomponentNode(name, "gltf-model", true, { src: uri, attribution, origin: originUri });
   }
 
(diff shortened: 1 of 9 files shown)


src/client/editor/Editor.js
```diff
@@ -1810,16 +1810,23 @@ export default class Editor {
     await this.exportScene(glbUri, true);
     const { id: glbId, token: glbToken } = await this.project.uploadAndDelete(glbUri);
 
-    const res = await this.project.createOrUpdateScene(
+    const sceneFileUri = this.sceneInfo.uri;
+    const { id: sceneFileId, token: sceneFileToken } = await this.project.uploadAndDelete(sceneFileUri);
```
brianpeiris: This is kinda dangerous because `uploadAndDelete` will actually delete the scene file off the disk. Usually this is fine because the code in `onPublish` will then immediately save the file, but if `createOrUpdateScene` fails, you will lose your scene file.

gfodor: Ohhh I saw a file disappear while debugging and thought I was crazy!

gfodor: nice catch!

PR #286, opened 2018-09-22
Export a preview camera into the GLTF

gfodor:
This PR exports a camera into the GLTF for the preview on the scene page. For now, the camera is set up as taking on the transform from the screenshot's camera.

### DIFF
--- a/src/client/editor/Editor.js
+++ b/src/client/editor/Editor.js
@@ -886,6 +886,16 @@ export default class Editor {
     const scene = this.scene;
     const clonedScene = scene.clone();
 
+    // Add a preview camera to the exported GLB if there is a transform in the metadata.
+    const { previewCameraTransform } = this.getSceneMetadata();
+
+    if (previewCameraTransform) {
+      const previewCamera = this.DEFAULT_CAMERA.clone();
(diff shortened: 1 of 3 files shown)


gfodor:
closes #280

src/client/editor/Editor.js
```diff
@@ -886,6 +886,16 @@ export default class Editor {
     const scene = this.scene;
     const clonedScene = scene.clone();
 
+    // Add a preview camera to the exported GLB if there is a transform in the metadata.
+    const { previewCameraTransform } = this.getSceneMetadata();
+
+    if (previewCameraTransform) {
+      const previewCamera = this.DEFAULT_CAMERA.clone();
+      previewCamera.name = "Preview Camera";
```
brianpeiris: Would it make sense to tag the camera with something more unique? Like setting `previewCamera.userData.spokePreviewCamera=true` or even just `previewCamera.name = "spoke-preview-camera"`?

gfodor: yeah that makes sense -- I was just winging it a bit here since we don't really know how, if at all, we will care about 'preview camera' as a concept. i'll change the name

PR #289, opened 2018-09-25
Packaging

mquander:
Now you can do `npm run binaries` and it will generate executables for Windows, Mac, and Linux, plus archives of that executable + the recast Node native module and stick them in the /release directory.


mquander:
By the way, I don't know what you want to do with this environment variable file, but personally I don't understand why both the server and client code need to know how to talk to Reticulum. Why not just the client? (I suggest this since all functionality by default should be on the client, so that we can run a hosted version without the server.) Maybe there is some reason I'm not seeing.

brianpeiris:
Changes look good and I've confirmed the linux executable works (yay!) but the windows executable failed for me:

image

brianpeiris:
I primarily used the environment variable for testing spoke against dev.ret, so it's mostly a development thing. It's fine if we make it a command line arg, or leave it hard-coded as it is NBD.

There are couple of reasons we have the node server communicate with reticulum. For the `import` function, it made more sense to have all that code live in the node server since the node server will eventually want to fetch and save the imported media to disk. For that and the other functions, we don't have to open up our hubs.mozilla.com CORS headers to allow localhost or whatever the user wants to call their local machine.

PR #293, opened 2018-10-03
Add client-side wasm version of Recast

mquander:
The packaged recast.js and recast.wasm are the build output of https://github.com/mquander/recastCLI.js. (The functions we use are exported from [here](https://github.com/mquander/recastCLI.js/blob/wasm/main/port.cpp).) Building it requires the Emscripten toolchain, so it seems inappropriate to include it in the build process here.

This seems to work great for small test scenes -- we should check it on larger scenes to verify its performance and correctness.


mquander:
If we want, we can remove the weird string serialization/deserialization step that we use to get the navigation mesh output now, since we're more in control of how we use Recast. I don't know how big our navigation meshes tend to get so I don't know if it matters at all. We would probably want to do it if we brought this functionality into Hubs, at least.

mquander:
I changed the build process to just output an executable now, since that's all that needs to ship. Let me know if you think outputting a zip file is still useful for some reason.

mquander:
I tested this with the Atrium environment and a Sketchfab model, and it seems fast and indistinguishable from the previous output.

donmccurdy:
This is awesome, great work @mquander! I'd wanted to do this in the browser originally but wasn't able to get the WASM build working. Nice to see it's only ~300kb in WASM too.

> If we want, we can remove the weird string serialization/deserialization step that we use to get the navigation mesh output now...

Have you looked at all into outputing Recast's own navmesh format(s), rather than the raw navmesh geometry? Or compiling `Detour` to WASM separately? At some point down the road I'd like to add more of Recast/Detour's features to `three-pathfinding`, or replace it entirely if Detour is simply better. That would provide features like moving obstacles, off-mesh links (doors), crowds, and more.

donmccurdy:
Also, if I can persuade you to put this WASM build of Recast on NPM (is that even possible?) I'd love to use it in https://github.com/donmccurdy/aframe-inspector-plugin-recast. 😇

mquander:
You can put it there if you want -- I don't endorse it as a fully-baked thing or a thing I expect to maintain, though, so I would rather not put it there. (We aren't sure if this functionality is going to wind up in the Spoke client, or in our server, or in Hubs, or what, all the work should probably be happening in a web worker in the JS side, and the surface area of Recast we might want to expose will change over time.)

I didn't consider any of that stuff because our environments are small enough that we haven't experienced any performance problems with the implementation as it stands. It all sounds like a bunch of good ideas that should be made though. I expect it's probably realistic to get the rest of Recast and Detour compiled to WebAssembly since it was easy to do this part.

PR #309, opened 2018-10-03
Support for importing/exporting animated gltf-models

### LINKED ISSUE #230: Decide on what to do about animations for v1
I think the ideal scenario is one where on each GLTF model you can choose one animation clip to play.

robertlong:
Closes #230


src/client/componentTypeMappings.js
```diff
@@ -8,27 +8,36 @@ import FileInput from "./ui/inputs/FileInput";
 import NumericInput from "./ui/inputs/NumericInput";
 import StringInput from "./ui/inputs/StringInput";
 import Vector3Input from "./ui/inputs/Vector3Input";
+import SelectInput from "./ui/inputs/SelectInput";
 
 /* eslint react/display-name: 0 */
 const componentTypeMappings = new Map([
-  [types.color, (value, isValid, onChange) => <ColorInput value={value} onChange={onChange} />],
-  [types.string, (value, isValid, onChange) => <StringInput value={value} onChange={onChange} />],
+  [types.color, (component, prop, value, isValid, onChange) => <ColorInput value={value} onChange={onChange} />],
```
gfodor: minor nit, but i'd call those `props`

src/client/editor/Editor.js
```diff
@@ -1011,6 +1013,33 @@ export default class Editor {
...
+        if (gltfRoot && clipName !== null) {
+          const animation = gltfRoot.animations.find(a => a.name === clipName);
+
+          if (animation) {
+            animations.push(animation);
+          } else {
+            throw new Error(`Animation for clip "${clipName}" not found.`);
+          }
+        } else if (loopAnimationComponent) {
+          const index = object.userData._components.findIndex(
```
gfodor: this seems like it should be broken out into a function in the spirit of the other component management APIs

src/client/editor/Editor.js
```diff
@@ -1390,6 +1420,16 @@ export default class Editor {
 
       object.add(scene);
       component._object = scene;
+
+      const animations = scene.animations;
+      const loopAnimationComponent = LoopAnimationComponent.getComponent(object);
```
gfodor: there's a `hasComponent` that can be used here and in other places in this PR

src/client/componentTypeMappings.js
robertlong: Yeah `prop` is a bit unclear. I think it should be `propDef` or `propSchema` though. I's a single entry from the component's schema.

src/client/editor/Editor.js
robertlong: We need to fix this in general. We should not have to have two different `getComponent` methods just because we clone a scene on export. I think we should move to real component classes in the future that are not stored in userData. Furthermore we should stop using three's clone method on object3D's. It doesn't work properly for skinned meshes and we can implement our own method of cloning components. I'll refactor this code after the beta launch.

PR #312, opened 2018-10-04
Update info mechanism, update notification and required update dialog

brianpeiris:


src/server/index.js
```diff
@@ -520,6 +523,78 @@ module.exports = async function startServer(options) {
...
+  };
+  function getDownloadUrlForCurrentPlatform(assets) {
+    const assetPlatform = nodePlatformToAssetPlatform[process.platform];
+    return assets.find(asset => asset.name.includes(assetPlatform)).downloadUrl;
+  }
+
+  const updateInfoTimeout = 2000;
+
+  async function getLatestRelease() {
+    const token = "de8cbfb4cc0281c7b731c891df431016c29b0ace";
```
brianpeiris: Github's graphql API requires a token, so I generated a personal access token with no special scopes.

src/server/index.js
```diff
@@ -520,6 +523,78 @@ module.exports = async function startServer(options) {
...
+    const result = await fetch("https://api.github.com/graphql", {
+      timeout: updateInfoTimeout,
+      method: "POST",
+      headers: { authorization: `bearer ${token}` },
+      body: JSON.stringify({
+        query: `
+          {
+            repository(owner: "mozillareality", name: "spoke") {
```
gfodor: do we need to include a order by here or something? I'm not that familiar with this API or graphql in general

gfodor: yeah looks like there is a way to order them, i'm guessing the default is probably OK but probably should confirm https://developer.github.com/v4/enum/releaseorderfield/

gfodor: it might be release date ascending by default, so we probably need to address this

src/server/index.js
```diff
@@ -520,6 +523,78 @@ module.exports = async function startServer(options) {
...
+    return {
+      version: release.tag.name,
+      downloadUrl: getDownloadUrlForCurrentPlatform(release.releaseAssets.nodes)
+    };
+  }
+
+  router.get("/api/update_info", koaBody(), async ctx => {
+    try {
+      // This endpoint doesn't exist yet but we query it for future use.
+      const configEndpoint = `https://${reticulumServer}/api/v1/config`;
```
gfodor: to be a bit pedantic, this should probably be `/api/v1/config/spoke`

src/client/ui/dialogs/UpdateRequiredDialog.js
```diff
@@ -0,0 +1,30 @@
...
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    return (
+      <div className={styles.dialogContainer}>
+        <Header title="Update Required" />
+        <div className={styles.updateRequiredContainer}>
+          This version of Spoke is no longer supported.<br />
```
gfodor: I think these forced updates will be more about bugs/security, so we can probably drop this sentence

src/client/ui/menus/ToolBar.js
```diff
@@ -150,9 +159,25 @@ export default class ToolBar extends Component {
...
+    const { updateAvailable, updateRequired, latestVersion, downloadUrl } = this.props.editor.updateInfo;
+    const showUpdateNotification = updateAvailable && !updateRequired && !this.state.updateNotificationDismissed;
     return (
       <div className={styles.toolbar}>
         <div className={styles.logo}>spoke</div>
+        {showUpdateNotification && (
+          <div className={styles.notification}>
+            <div className={styles.message}>
+              {`Spoke ${latestVersion} is available. `}
+              <Button className={styles.download} href={downloadUrl}>
```
gfodor: should we target blank this, so they don't lose work?

src/server/index.js
gfodor: actually this is non-pedantic since i guess this URL actually matters :)

brianpeiris: `/api/v1/configs/spoke`?

src/client/ui/menus/ToolBar.js
brianpeiris: Firefox just starts the download. Will check Chrome.

brianpeiris: Chrome does the same. The response from the URL that github provides includes a content disposition header, so browsers just download the release.

PR #318, opened 2018-10-05
Add viewport status bar with rendering stats

mquander:
Seems nice to have. Looks like this:

stats


mquander:
Over here you can see the other information which is equally easy to get: https://threejs.org/docs/#api/en/renderers/WebGLRenderer.info

mquander:
On Slack it sounded like people thought that material count was a better proxy for efficiency than draw calls, so I changed it.

PR #324, opened 2018-10-09
Lock button for uniform scaling

### LINKED ISSUE #250: Uniform Scale in PropertiesPanel
This would be very useful for Sketchfab models which often in the cm scale

robertlong:
Fixes #250

### DIFF
--- a/src/client/ui/inputs/Vector3Input.js
+++ b/src/client/ui/inputs/Vector3Input.js
@@ -1,34 +1,89 @@
-import React from "react";
+import React, { Component } from "react";
 import PropTypes from "prop-types";
 import styles from "./Vector3Input.scss";
+import classNames from "classnames";
 import NumericInput from "./NumericInput";
 
-export default function Vector3Input({ value, onChange, ...rest }) {
-  const vx = value ? value.x : 0;
-  const vy = value ? value.y : 0;
-  const vz = value ? value.z : 0;
-  return (
-    <div className={styles.inputGroup}>
-      <div className={styles.label}>X:</div>
-      <NumericInput value={vx} onChange={x => onChange({ x, y: vy, z: vz })} {...rest} />
-      <div className={styles.label}>Y:</div>
-      <NumericInput value={vy} onChange={y => onChange({ x: vx, y, z: vz })} {...rest} />
-      <div className={styles.label}>Z:</div>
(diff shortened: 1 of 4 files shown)


src/client/ui/inputs/Vector3Input.js
```diff
@@ -1,34 +1,89 @@
...
+    const { uniformEnabled } = this.state;
+    const vx = value ? value.x : 0;
+    const vy = value ? value.y : 0;
+    const vz = value ? value.z : 0;
+
+    const uniformClassName = classNames(styles.uniformButton, uniformEnabled ? styles.uniformEnabled : undefined);
+
+    return (
+      <div className={styles.inputGroup}>
+        {uniformScaling && <i className={uniformClassName} onClick={this.onToggleUniform} />}
```
brianpeiris: I would make this a `<button>` or even a `<input type="checkbox">` instead of an `<i>` for semantic and accessibility reasons.

src/client/ui/inputs/Vector3Input.scss
```diff
@@ -14,3 +15,29 @@
...
+  display: flex;
+  align-items: center;
+  cursor: pointer;
+
+  &:before {
+    color: $text2;
+    font-size: 10px;
+    padding: 0 6px;
+    @extend .fas;
+    content: fa-content($fa-var-unlock);
```
brianpeiris: I think this is better represented with a chain link icon. A lock implies that the values cannot change. https://fontawesome.com/icons/link?style=solid

src/client/ui/inputs/Vector3Input.js
brianpeiris: Maybe also add a `title=""` for a tooltip?

robertlong: But everything in React is a div 😄 Yeah, very true. I'll go with a checkbox and title. It's a bit harder to style but better for everyone.

src/client/ui/inputs/Vector3Input.scss
robertlong: Sure, I have no strong opinion. I've seen the lock in Sketch and the link in Figma. Let's go with the link.

gfodor:
awesome!

PR #327, opened 2018-10-10
Fix SpokeSpotLightHelper orientation

### LINKED ISSUE #297: Light gizmos don't match what light actually is doing
The light gizmos, especially with spotlights, aren't aligned to the light positions and orientations, making them very difficult to manipulate.

robertlong:
Fixes #297

### DIFF
--- a/src/client/editor/helpers/SpokeSpotLightHelper.js
+++ b/src/client/editor/helpers/SpokeSpotLightHelper.js
@@ -10,9 +10,6 @@ export default class SpokeSpotLightHelper extends THREE.Object3D {
     this.matrix = light.matrixWorld;
     this.matrixAutoUpdate = false;
 
-    this.vector = new THREE.Vector3();
-    this.vector2 = new THREE.Vector3();
-
     this.color = color;
 
     const geometry = new THREE.BufferGeometry();
@@ -54,12 +51,6 @@ export default class SpokeSpotLightHelper extends THREE.Object3D {
 
     this.outerCone.scale.set(outerConeWidth, outerConeWidth, coneLength);
     this.innerCone.scale.set(innerConeWidth, innerConeWidth, coneLength);
-
-    this.vector.setFromMatrixPosition(this.light.matrixWorld);
-    this.vector2.setFromMatrixPosition(this.light.target.matrixWorld);
-
-    this.outerCone.lookAt(this.vector2.sub(this.vector));
-    this.innerCone.lookAt(this.vector2.sub(this.vector));
 
     if (this.color !== undefined) {
       this.outerCone.material.color.set(this.color);


robertlong:
I'm going to expand this PR to include the rest of the lighting helper changes.

src/client/editor/helpers/SpokePointLightHelper.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    this.add(this.lightDistanceHelper);
+
+    this.update();
+  }
+
+  dispose() {
+    this.geometry.dispose();
+    this.material.dispose();
```
brianpeiris: Probably also want to dispose the `lightDistanceHelper` here.

src/client/editor/helpers/SpokePointLightHelper.js
```diff
@@ -0,0 +1,63 @@
...
+
+    this.lightDistanceHelper = new THREE.Mesh(distanceGeometry, distanceMaterial);
+
+    const d = light.distance;
+
+    if (d === 0.0) {
+      this.lightDistanceHelper.visible = false;
+    } else {
+      this.lightDistanceHelper.scale.set(d, d, d);
+    }
```
brianpeiris: Can delete this if block I think, since it's going to happen in `update` anyway

src/client/editor/helpers/SpokePointLightHelper.js
```diff
@@ -0,0 +1,63 @@
...
+
+    this.light = light;
+    this.light.updateMatrixWorld();
+
+    this.color = color;
+
+    this.matrix = this.light.matrixWorld;
+    this.matrixAutoUpdate = false;
+
+    const distanceGeometry = new THREE.IcosahedronGeometry(1, 2);
```
brianpeiris: Use IcosahedronBufferGeometry?

src/client/editor/helpers/SpokePointLightHelper.js
```diff
@@ -0,0 +1,63 @@
...
+    const distanceMaterial = new THREE.MeshBasicMaterial({
+      color: color,
+      fog: false,
+      wireframe: true,
+      opacity: 0.1,
+      transparent: true
+    });
```
brianpeiris: I'm not sure about having this visible at all times when a distance is set. Might get annoying. Ideally the range helper would only be visible while the point light is selected? Would be nice for a future version.

robertlong: Maybe all the light helpers should function this way?

brianpeiris: Hmm, I think you want some simple gizmo so that you can actually see that the lights exist in the scene, kinda like Unity's icon gizmo and then have them expand to the full helper representation when they are selected.

robertlong: Ok, yeah. Let's do that soon.

PR #338, opened 2018-10-15
Tweak nav mesh parameters

robertlong:
Currently the nav mesh generation uses a cell size that is based on the area of the bounding box of the environment. This produces a cell size that can be too large and result in doorways and halls that are blocked. I propose that we set the cell size to a constant 0.25 meters. In my testing this generates a correct navmesh for correctly sized doorways and halls.

If reducing the cell size results in navmesh generation times that are too long in larger environments, I would suggest that we reduce the maximum size of the environment. I believe it is more crucial for us to have smaller environments where the expected areas are traversable than large wide open areas.

It also seems to make sense to expose some of these navmesh settings at some point. Cell size, cell height, agent max slope, region min size, etc. are all things that could be adjusted based on the features of your environment. Once we have the time to introduce the UI I'd like to expose those settings.

### DIFF
--- a/src/client/editor/Editor.js
+++ b/src/client/editor/Editor.js
@@ -864,8 +864,13 @@ export default class Editor {
       const box = new THREE.Box3().setFromBufferAttribute(finalGeo.attributes.position);
       const size = new THREE.Vector3();
       box.getSize(size);
-      if (Math.max(size.x, size.y, size.z) > 2000) {
-        throw new Error(`Scene is too large (${size.x} x ${size.y} x ${size.z}) to generate a nav mesh.`);
+
+      if (size.x * size.y * size.z > 1000000) {
+        throw new Error(
+          `Scene is too large (${size.x} x ${size.y} x ${
+            size.z
+          }) to generate a nav mesh. Maximum volume is 1,000,000 cubic meters.`
+        );
       }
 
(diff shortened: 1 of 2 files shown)


brianpeiris:
Here are the models I originally tested against when tuning the nav mesh parameters:

https://sketchfab.com/models/7846cccb4c794860a34d8ab5adabff1e
https://sketchfab.com/models/485266e0d43e46a7bc7850d3d957571a
https://sketchfab.com/models/b123b8c6f41f41ef8b3b8f98975eb109
https://sketchfab.com/models/326af8ee3e814812993e9cc15732234c

arpu:
i have some problems to generate a navmesh with stairs 

https://drive.google.com/open?id=11b55oL6n65ThkKuUA_-o0wjehdozkXoi

robertlong:
@arpu is this what you are seeing?

screen shot 2018-10-17 at 9 30 03 am

I'm not certain we can achieve much more detail than this with our current navmesh generation settings. I'm hoping that we can move away from using navmeshes altogether by switching to mesh colliders and an updated character controller for Hubs in the near future.

arpu:
yeah exactly the stairs are not generated 

ok thx for the information

robertlong:
Closing. We added the option to exclude meshes from the nav mesh.

PR #339, opened 2018-10-16
Combine meshes with identical materials on export

robertlong:
Reduces the total number of meshes, materials, and textures in final exported model.

### DIFF
--- a/src/client/editor/Editor.js
+++ b/src/client/editor/Editor.js
@@ -20,10 +20,10 @@ import {
   StaticModes,
   getStaticMode,
   computeStaticMode,
-  computeAndSetStaticModes,
   getOriginalStaticMode,
   setOriginalStaticMode
 } from "./StaticMode";
+import MeshCombinationGroup from "./MeshCombinationGroup";
 import { generateNavMesh } from "../utils/navmesh";
 import { getUrlFilename } from "../utils/url-path";
 import { textureCache, gltfCache } from "./caches";
@@ -39,7 +39,6 @@ import SpokePointLightHelper from "./helpers/SpokePointLightHelper";
 import { last } from "../utils";
 import absoluteToRelativeURL from "./utils/absoluteToRelativeURL";
 import addChildAtIndex from "./utils/addChildAtIndex";
-import shallowEquals from "./utils/shallowEquals";
 import sortEntities from "./utils/sortEntities";
 import cloneObject3D from "./utils/cloneObject3D";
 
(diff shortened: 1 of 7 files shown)


src/client/editor/MeshCombinationGroup.js
```diff
@@ -0,0 +1,165 @@
...
+    return a.src === b.src || getBase64Image(a) === getBase64Image(b);
+  }
+
+  return a === b;
+}
+
+export function compareTextures(a, b) {
+  if (a && b) {
+    return (
+      compareImages(a, b) &&
```
gfodor: slight perf improvement might be to re-order these to do the expensive operation last

src/client/editor/Editor.js
```diff
@@ -1419,11 +1350,17 @@ export default class Editor {
         object.remove(component._object);
       }
 
+      const animations = scene.animations;
+
+      if (animations.length > 0) {
+        setStaticMode(scene, StaticModes.Dynamic);
+      } else {
+        setStaticMode(scene, StaticModes.Static);
+      }
```
gfodor: does this mean that if there are any animations in the scene, we skip the combination for the whole scene?

src/client/editor/MeshCombinationGroup.js
```diff
@@ -0,0 +1,165 @@
...
+  constructor(initialObject) {
+    if (!initialObject.isMesh) {
+      throw new Error("MeshCombinationGroup must be initialized with a THREE.Mesh.");
+    }
+
+    this.initialObject = initialObject;
+    this.meshes = [initialObject];
+  }
+
+  tryAdd(object) {
```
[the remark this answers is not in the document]
gfodor: yeah I think all of them except the static one could/should have underscores

gfodor:
BTW, nice implementation, easy to read and understand!

src/client/editor/MeshCombinationGroup.js
```diff
@@ -0,0 +1,165 @@
...
+  // guess the original format, but be aware the using "image/jpg"
+  // will re-encode the image.
+  const dataURL = canvas.toDataURL("image/png");
+
+  return dataURL.replace(/^data:image\/(png|jpg);base64,/, "");
+}
+
+export function compareImages(a, b) {
+  if (a && b) {
+    return a.src === b.src || getBase64Image(a) === getBase64Image(b);
```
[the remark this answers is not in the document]
gfodor: actually caching these may actually be something to be cautious about -- it would (I think) cause a large relative increase in memory pressure vs the way it is now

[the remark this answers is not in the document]
gfodor: you could cache the sha of the base64 image instead

src/client/editor/Editor.js
robertlong: The whole loaded glTF scene, yes. I'm not sure how to make things safe with how ThreeJS manages animation paths.

gfodor: so if I have two models in my spoke scene, one has no animations, and the other does, the one without animations will still be combined? i guess that's OK -- seems like a tough tradeoff though.

robertlong: I'm confused, is this different from what you had in mind? Maybe you are confusing glTF scene and Spoke scene here? I'm not able to safely combine the ThreeJS object3Ds that are loaded from a glTF model with animations. It would require determining which object3Ds are included in the animation and should be left as dynamic.

PR #340, opened 2018-10-17
Migrate glTF namespace HUBS_components -> MOZ_hubs_components

robertlong:
We currently do not have the `HUBS` glTF vendor prefix registered and should probably not be publishing assets with it. Instead of registering a new vendor prefix for Hubs, which at this point seems more like a product than a vendor, I propose we use the `MOZ_hubs_` prefix for all Hubs specific extensions.

If you'd like to read more on this topic see this glTF issue: https://github.com/KhronosGroup/glTF/issues/1387

### DIFF
--- a/src/client/editor/Editor.js
+++ b/src/client/editor/Editor.js
@@ -1047,8 +1047,8 @@ export default class Editor {
       if (userData.gltfExtensions === undefined) {
         userData.gltfExtensions = {};
       }
-      if (userData.gltfExtensions.HUBS_components === undefined) {
-        userData.gltfExtensions.HUBS_components = {};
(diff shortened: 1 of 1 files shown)


brianpeiris:
LGTM. You'll have to to change the Hubs code as well of course:
https://github.com/mozilla/hubs/blob/master/src/components/gltf-model-plus.js#L100

And presumably you'll want to change it in a way that preserves currently existing scenes that use `HUBS_components`?

robertlong:
Cool, I'll make the PR for Hubs as well.

robertlong:
This is now outdated. I'll revisit this soon.

PR #347, opened 2018-10-19
Use websockets to report upload progress and show progress status during publish

brianpeiris:
This fixes a timeout bug on slow/average broadband upload speeds and also gives you publish and upload progress updates.


brianpeiris:
Thanks. You're right about failed uploads leaving files behind. Since we auto-generate all of the uploaded files anyway, I think it makes sense to delete them on failure as well, so I'll make that change.

PR #354, opened 2018-10-25
Merge meshes in animated glTF models

robertlong:
Merges meshes in animated models by marking animated nodes as dynamic. Results in significantly less draw calls in environments using models like this: https://sketchfab.com/models/7984bd6024db4b89928ddeb625eebac0

screen shot 2018-10-25 at 2 10 00 pm


gfodor:
nice work! 🎉

PR #358, opened 2018-11-02
Feature/structured attributions

gfodor:
Updates Spoke to POST structured attribution data about content, requires https://github.com/mozilla/reticulum/pull/83

Scene attribution data will now include JSON data for each piece of imported content so it can be properly rendered as links, etc, as in https://github.com/mozilla/hubs/pull/664

### DIFF
--- a/src/client/editor/Editor.js
+++ b/src/client/editor/Editor.js
@@ -1224,7 +1224,11 @@ export default class Editor {
 
   async addGLTFModelNode(name, uri, originUri) {
     const attribution = await this.project.getImportAttribution(uri);
-    this.addUnicomponentNode(name, "gltf-model", { src: uri, attribution, origin: originUri });
+    this.addUnicomponentNode(name, "gltf-model", {
+      src: uri,
+      attribution: JSON.stringify(attribution),
+      origin: originUri
+    });
   }
 
   async importGLTFIntoModelNode(url) {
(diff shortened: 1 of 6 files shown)


src/client/editor/Editor.js
```diff
@@ -1858,33 +1862,53 @@ export default class Editor {
     this.history.redo();
   }
 
-  getSceneAttribution() {
-    const attributions = new Set();
+  getSceneContentAttributions() {
+    const contentAttributions = [];
```
robertlong: Nit: You can probably just make `contentAttributions` a Set instead of having both an array and a set. If you want to return an array so can use all the array methods, you can use `return Array.from(contentAttributions)`

src/client/editor/Editor.js
```diff
@@ -1224,7 +1224,11 @@ export default class Editor {
 
   async addGLTFModelNode(name, uri, originUri) {
     const attribution = await this.project.getImportAttribution(uri);
-    this.addUnicomponentNode(name, "gltf-model", { src: uri, attribution, origin: originUri });
+    this.addUnicomponentNode(name, "gltf-model", {
+      src: uri,
+      attribution: JSON.stringify(attribution),
```
robertlong: Does this work for the legacy attribution? Doesn't `getImportAttribution` return an object?

robertlong: Oh I see you parse it later. Interesting.

PR #359, opened 2018-11-05
Project Restructure

robertlong:
- Reorganize the UI folder so that components are grouped logically.
- Further separate editor, UI and backend interface code.
  - Remaining tasks include:
    - Figure out a better way of exposing AuthenticationError to the UI
    - AddNodeActionButtons still contains import references to Node classes. The add node UI should be dynamically generated from the map of Node -> NodeEditor which is now a top level API.
- Move url helpers into project so they can be overridden depending on the backend.


src/client/editor/caches/GLTFCache.js
```diff
@@ -72,7 +39,7 @@ class GLTFCache extends Cache {
             if (prop instanceof THREE.Texture) {
               if (prop.image.src) {
                 const absoluteTextureURL = new URL(prop.image.src, window.location).href;
-                textureCache._cache.set(absoluteTextureURL, Promise.resolve(prop));
+                this.textureCache._cache.set(absoluteTextureURL, Promise.resolve(prop));
```
gfodor: i thought our convention was `_` are private members?

src/client/editor/commands/AddObjectCommand.js
```diff
@@ -24,12 +24,18 @@ export default class AddObjectCommand extends Command {
...
+
+    this.editor.signals.objectAdded.dispatch(this.object);
+    this.editor.signals.sceneGraphChanged.dispatch();
     this.editor.select(this.object);
   }
 
   undo() {
     this.editor.removeObject(this.object);
+
+    this.editor.signals.objectAdded.dispatch(this.object);
```
gfodor: dispatch of `objectAdded` intentional?

src/client/editor/nodes/AmbientLightNode.js
```diff
@@ -0,0 +1,52 @@
+import * as THREE from "three";
+import EditorNodeMixin from "./EditorNodeMixin";
+import serializeColor from "../utils/serializeColor";
+
+export default class AmbientLightNode extends EditorNodeMixin(THREE.AmbientLight) {
+  static legacyComponentName = "ambient-light";
+
+  static nodeName = "Ambient Light";
```
gfodor: I think we'll probably be living with `legacyComponentName` for a while, perhaps we should just call these `serializedName` and `displayName`?

gfodor: or I guess since we use this for the GLTF component (and will for the forseeable future) perhaps just `componentName`

src/client/editor/nodes/DirectionalLightNode.js
```diff
@@ -0,0 +1,90 @@
...
+          this.add(clonedChild);
+        }
+      }
+    }
+
+    return this;
+  }
+
+  serialize() {
+    const json = super.serialize();
```
gfodor: these `serialize` and `prepareForExport` routines don't seem completely DRY. doesn't matter that much but might be worth extracting functions

src/client/editor/commands/AddObjectCommand.js
robertlong: Yeah, this is consistent with the other commands. I'm not certain if these events belong in commands but this is more consistent with the others.

src/client/editor/caches/GLTFCache.js
robertlong: Yeah, I probably should make cache public then. Thanks

src/client/editor/nodes/AmbientLightNode.js
robertlong: I think we have a good opportunity to move to a better serialization format when we moved to a hosted version of Spoke, so this probably won't stick around long. Node name was display name before, Idk why I changed it TBH. I'll change it back.

src/client/editor/nodes/AmbientLightNode.js
```diff
@@ -0,0 +1,52 @@
...
+    node.intensity = intensity;
+
+    return node;
+  }
+
+  serialize() {
+    const json = super.serialize();
+
+    json.components.push({
+      name: "ambient-light",
```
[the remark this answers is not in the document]
robertlong: Yup, seems like a good idea.

PR #362, opened 2018-11-06
Fix Cloning TransformControls

robertlong:
- Moves three.js back to a vendor file so that it can be required correctly.
- Introduces `SpokeTransformControls` which has the `.clone()` method overridden so that it returns an Object3D and avoids a bug with multiple instances of TransformControls.

### DIFF
--- a/src/client/editor/Editor.js
+++ b/src/client/editor/Editor.js
@@ -1,6 +1,6 @@
 import signals from "signals";
 import uuid from "uuid/v4";
-import * as THREE from "three";
+import THREE from "../vendor/three";
 import History from "./History";
 import Viewport from "./Viewport";
 
--- a/src/client/editor/MeshCombinationGroup.js
+++ b/src/client/editor/MeshCombinationGroup.js
@@ -1,4 +1,4 @@
-import * as THREE from "three";
+import THREE from "../vendor/three";
 import { isStatic } from "./StaticMode";
 import asyncTraverse from "./utils/asyncTraverse";
 import keysEqual from "./utils/keysEqual";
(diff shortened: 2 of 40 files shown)


gfodor:
Why does transformcontrols end up getting cloned in the first place? this makes me concerned there may be other non-scene oriented objects being inadvertently cloned with unintended side effects

robertlong:
I removed the helper scene from Spoke and we clone the whole scene on export. The helper scene created some complexity around what helpers actually existed in the helper scene vs the actual scene. For example light helpers, which were children of nodes in the scene graph and were not part of the helper scene. Removing the helper scene as a concept lets us focus on one scene graph and I think that will help as things like rendering order and more complex helpers come into this.

However, as a side effect we now clone objects that were in the helper scene. I think all objects should be able to be cloned. The fact that TransformControls can't is a design flaw.

When exporting we [ignore all objects on the scene that aren't nodes](https://github.com/MozillaReality/Spoke/blob/master/src/client/editor/nodes/SceneNode.js#L105) and this constraint makes filtering out helper objects pretty easy and avoids all of these userData properties we were setting before.

robertlong:
Also I don't think `prepareForExport` is the ideal solution. Without changing how GLTFExporter works I don't see an easy way around it. It's a stand in until we can figure out how to treat glTF export as a first class citizen in Spoke.

gfodor:
OK that makes more sense. I would consider adding some clarifying comments above the `isNode` field to explicitly call out that designates it as a "Spoke-specific Node, which includes it as part of the set of scene objects to include when exporting"

gfodor:
or perhaps add a documenting function on nodes called `includeInExport` that just returns `isNode`, which would make it a bit more self-explanatory. dont' have a strong opinion on this but calling that out probably would help

robertlong:
Ok, I'll open a new PR to clarify that.

PR #364, opened 2018-11-06
Fix box collider component

robertlong:
The `box-collider` component in Hubs expects transform properties which I forgot to export.


robertlong:
This currently isn't working. Hubs still doesn't seem to be setting the box-collider transform properties correctly. Not sure why yet.

robertlong:
I think it actually does work I was just seeing an error when rotating box colliders due to the orientation being reused across multiple shape objects: https://github.com/mozilla/hubs/blob/master/src/gltf-component-mappings.js#L37

PR #384, opened 2018-11-16
Shadows

### LINKED ISSUE #124: Enable Shadows

robertlong:
- Adds checkboxes to the UI to cast / receive shadows 
- Sets the default shadow type to PCF Soft
- Sets the default shadow map size for the directional light to 1024 x 1024 vs the default 512x512

To Do:
- [x] Set Hub's default shadow type to PCF Soft and remove `scene-shadow` component. It doesn't currently do anything because scene shadow type cannot be changed at runtime.
- [x] Set Hub's directional shadow map size to 512 x 512.
- [x] Move the directional light's shadow camera from using a fixed frustum to a frustum whose size is based on all objects marked with `castShadow` in the scene.

Fixes #124


src/client/editor/nodes/SceneNode.js
```diff
@@ -116,6 +116,14 @@ export default class SceneNode extends EditorNodeMixin(THREE.Scene) {
...
       node.prepareForExport();
     }
+
+    this.userData.gltfExtensions = {
+      HUBS_components: {
+        "scene-shadow": {
+          type: "pcfsoft"
+        }
+      }
+    };
```
brianpeiris: Don't we remove scene-shadow in https://github.com/mozilla/hubs/pull/755?

PR #386, opened 2018-11-22
Load Models via Media API and CORS Proxy

robertlong:
- Adds media utils.
- Refactor node API to expose editor.
- Stop importing models and caching on disk in favor of loading them from the canonical URL at runtime and using the browser cache. This makes spoke scenes more portable and makes it possible to download Spoke scenes and remix them.
- Regenerates the nav mesh if it cant be loaded from disk.

To Do:
- [x] Add header to caching assets loaded via cors-proxy.

### DIFF
--- a/src/client/editor/nodes/ModelNode.js
+++ b/src/client/editor/nodes/ModelNode.js
@@ -49,8 +49,20 @@ export default class ModelNode extends EditorNodeMixin(Model) {
     super(editor);
     this.attribution = null;
     this.includeInFloorPlan = true;
+    this._canonicalUrl = null;
   }
 
+  // Overrides Model's src property and stores the original (non-resolved) url.
+  get src() {
+    return this._canonicalUrl;
+  }
+
+  // When getters are overridden you must also override the setter.
+  set src(value) {
+    this.load(value).catch(console.error);
+  }
+
+  // Overrides Model's loadGLTF method and uses the Editor's gltf cache.
   async loadGLTF(src) {
     const gltf = await this.editor.gltfCache.get(src);
 
@@ -66,14 +78,16 @@ export default class ModelNode extends EditorNodeMixin(Model) {
     return gltf;
   }
 
+  // Overrides Model's load method and resolves the src url before loading.
   async load(src) {
-    this._originalSrc = src;
+    this._canonicalUrl = src;
 
     const { accessibleUrl, files } = await this.editor.project.resolveMedia(src);
 
     await super.load(accessibleUrl);
 
     if (files) {
+      // Revoke any object urls from the SketchfabZipLoader.
       for (const key in files) {
         URL.revokeObjectURL(files[key]);
       }
@@ -104,7 +118,7 @@ export default class ModelNode extends EditorNodeMixin(Model) {
     json.components.push({
       name: "gltf-model",
       props: {
-        src: absoluteToRelativeURL(sceneUri, this._originalSrc),
+        src: absoluteToRelativeURL(sceneUri, this._canonicalUrl),
         attribution: this.attribution,
         includeInFloorPlan: this.includeInFloorPlan
       }


src/client/editor/objects/Model.js
```diff
@@ -6,21 +6,39 @@ export default class Model extends THREE.Object3D {
...
+
+  set src(value) {
+    this.load(value).catch(console.error);
+  }
+
+  loadGLTF(src) {
+    return new Promise((resolve, reject) => {
+      new THREE.GLTFLoader().load(src, resolve, null, reject);
+    });
+  }
```
brianpeiris: Hmm, a bit confusing to tell if `set src` and `loadGLTF` here are actually used. I'm pretty sure `loadGLTF` here never gets called, since we override it in ModelNode and we never call `Model.loadGLTF` directly, or indirectly.

src/client/ui/viewport/ViewportPanelContainer.js
```diff
@@ -29,94 +24,15 @@ class ViewportPanelContainer extends Component {
     this.props.editor.createViewport(this.canvasRef.current);
   }
 
-  onDropAsset = async item => {
```
brianpeiris: Removed support for dropping urls and files?

src/server/index.js
```diff
@@ -334,74 +321,63 @@ async function startServer(options) {
...
         method: "POST",
         headers: { "content-type": "application/json" },
-        body: JSON.stringify({ media: { url: origin, index: 0 } })
+        body: JSON.stringify(ctx.request.body)
       });
-
+      ctx.status = resp.status;
       if (resp.status !== 200) {
-        ctx.status = resp.status;
-        ctx.body = await resp.text();
```
brianpeiris: Might want to at least log `await resp.text()`. Same below.

src/client/api/media.js
```diff
@@ -0,0 +1,85 @@
...
+const resolveUrlCache = new Map();
+
+async function resolveUrl(url, index) {
+  const cacheKey = `${url}|${index}`;
+  if (resolveUrlCache.has(cacheKey)) return resolveUrlCache.get(cacheKey);
+  const resolved = await fetch("/api/media", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({ media: { url, index } })
+  }).then(r => r.json());
```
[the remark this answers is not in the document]
robertlong: Ok, I think I have a better solution. I'll also fix it in Hubs. https://github.com/mozilla/hubs/blob/master/src/utils/media-utils.js#L51

src/client/ui/viewport/ViewportPanelContainer.js
robertlong: Yeah, unfortunately it already wasn't really working. I plan to add proper support back in soon.

src/client/api/sketchfab-zip.worker.js
```diff
@@ -0,0 +1,33 @@
...
+  const fileBlobs = await Promise.all(files.map(f => f.async("blob")));
+  for (let i = 0; i < fileBlobs.length; i++) {
+    fileMap[files[i].name] = URL.createObjectURL(fileBlobs[i]);
+  }
+
+  const gltfJson = JSON.parse(await zip.file("scene.gltf").async("text"));
+  gltfJson.buffers && gltfJson.buffers.forEach(b => (b.uri = fileMap[b.uri]));
+  gltfJson.images && gltfJson.images.forEach(i => (i.uri = fileMap[i.uri]));
+
+  fileMap["scene.gtlf"] = URL.createObjectURL(new Blob([JSON.stringify(gltfJson, null, 2)], { type: "text/plain" }));
```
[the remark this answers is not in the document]
robertlong: Hmm. Yeah that's not great. I think I implemented a solution but wasn't able to record any change in memory usage. I don't know if I'm doing this correctly. Also I don't think Hubs is either: https://github.com/mozilla/hubs/blob/master/src/components/gltf-model-plus.js#L278 that line isn't being called with the actual object url, it's being called with the file name.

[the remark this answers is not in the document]
robertlong: @netpro2k

src/client/editor/objects/Model.js
robertlong: Neither are called. My only excuse is that Spoke objects are just Three.js Object3Ds. They should work correctly outside the editor and could be used in a future runtime project. Spoke's Nodes are tied to the editor and so they override those methods. There's not much code in those overridden methods so I don't see much harm in it. Ideally I'd like to see these objects split out into another library of useful ThreeJS objects.

brianpeiris: I would add some comments at least. It would help anyone else reading the spoke code.

PR #387, opened 2018-11-22
Image and Video Nodes

robertlong:
Merge #386 first. To be merged with https://github.com/mozilla/hubs/pull/783

To Do:
- [x] Implement image node in Spoke
- [x] Implement image component inflator in Hubs
- [x] Implement video node in Spoke
- [x] Implement video component inflator in Hubs


src/client/editor/nodes/ImageNode.js
```diff
@@ -0,0 +1,82 @@
...
+          src: this._canonicalUrl
+        },
+        networked: {
+          id: this.uuid
+        }
+      }
+    };
+
+    this.parent.add(replacementObject);
+    this.parent.remove(this);
```
netpro2k: Again probably not specific to this PR since I assume this is just a pattern we are using all around Spoke but its pretty confusing why the scene graph ends up getting manipulated during export. What is responsible for putting the editor node back in? 

Why would this not just `serialize` (or a something very similar, since I guess serialize/deserialize is for the editor format no the runtime format)? I assume this is also where all the "copy" issues you have talked about in the past comes from. Is this all just so that we don't have to modify the GLTF serialzier from Three? 

Seems like it could be done in a similar way to what we did on the loading side, by augmenting it with callbacks for extensions. Though I admittedly have not thought about it as much as you have so I may be over simplifying.

src/client/editor/nodes/FloorPlanNode.js
```diff
@@ -282,6 +279,19 @@ export default class FloorPlanNode extends EditorNodeMixin(FloorPlan) {
     return this;
   }
 
+  async loadNavMesh(src) {
```
netpro2k: Does this actually relate to the media stuff at all or did it just slip in?

src/client/editor/nodes/SpawnPointNode.js
```diff
@@ -11,8 +11,7 @@ export default class SpawnPointNode extends EditorNodeMixin(THREE.Object3D) {
 
   static async load() {
     const { scene } = await new Promise((resolve, reject) => {
-      const loader = new THREE.GLTFLoader();
-      loader.load(spawnPointModelUrl, resolve, null, reject);
+      new THREE.GLTFLoader().load(spawnPointModelUrl, resolve, null, reject);
```
netpro2k: Probably don't want to create a new loader every time. Also this doesnt use `this` so it should just bbe a function. Might also make sense to just load once it at startup and then clone, since right now you end up re-parsing the gltf for each spawn point.

src/client/editor/objects/Model.js
```diff
@@ -6,21 +6,39 @@ export default class Model extends THREE.Object3D {
...
+    return this._src;
+  }
+
+  set src(value) {
+    this.load(value).catch(console.error);
+  }
+
+  loadGLTF(src) {
+    return new Promise((resolve, reject) => {
+      new THREE.GLTFLoader().load(src, resolve, null, reject);
```
netpro2k: creating new loader every time, not using `this` so it can be a function

src/client/ui/properties/VideoNodeEditor.js
```diff
@@ -0,0 +1,128 @@
...
+
+export default class VideoNodeEditor extends Component {
+  static propTypes = {
+    editor: PropTypes.object,
+    node: PropTypes.object
+  };
+
+  static iconClassName = "fa-video";
+
+  setProp(propName) {
```
netpro2k: Very minor but the extra binding here isn't needed, and the name feels a bit misleading since this actually doesn't set a prop, it returns a function that does. I probably would have wrote this as a function outside the class like:
```
function propSetter(propName) {
  return function(value) {
    this.props.editor.setNodeProperty(this.props.node, propName, value);
  }
}
```

and used:
```
onChangeControls: propSetter("controls"),
```

src/client/api/media.js
```diff
@@ -0,0 +1,94 @@
+import SketchfabZipWorker from "./sketchfab-zip.worker.js";
```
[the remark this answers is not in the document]
robertlong: Hmm, media-utils has some external dependencies that I don't want to bring into Spoke. I think I'm going to keep them separate.

src/client/editor/nodes/ImageNode.js
robertlong: We clone before mutating the scene graph so we don't really have to worry about the mutations. I'd rather just modify GLTFExporter so we didn't have to do this, but there are currently no hooks to be able to accomplish that. It is a complicated API to design, maybe we should just fork the GLTFExporter and figure out where we should create API hooks.

robertlong: Cloning/copying are also needed for duplicating objects. So we still need those methods to work properly.

src/client/editor/nodes/FloorPlanNode.js
robertlong: This branch is based off of the `feature/farspark-model` branch. I'm surprised this is still showing up in the diff. https://github.com/MozillaReality/Spoke/pull/386/files#diff-3a1b4a86fe9a09ae9d0623b70ac64336R282

src/client/editor/nodes/SpawnPointNode.js
robertlong: This one is actually only loaded once at startup and then cloned. I'm going to keep it as is.

src/client/editor/objects/Model.js
robertlong: This method is overidden in ModelNode and we use the GLTFCache rather than the GLTFLoader there. There isn't much overhead to creating a new GLTFLoader and it doesn't need to be held onto after it is done loading. I actually think it might be better to create a loader per asset. There is some state that is associated with the currently loading model so you probably don't want to reuse it if you are loading in parallel.

src/client/ui/properties/VideoNodeEditor.js
robertlong: `this` will not be bound to the correct value. I think renaming it to `createPropSetter` is a good idea.

Maybe this is the cleanest implementation?
```
constructor(props) {
    super(props);
    const createPropSetter = propName => value => this.props.editor.setNodeProperty(this.props.node, propName, value);
    this.onChangeControls = createPropSetter("controls");
    this.onChangeAutoPlay = createPropSetter("autoPlay");
    this.onChangeLoop = createPropSetter("loop");
    this.onChangeAudioType = createPropSetter("audioType");
    this.onChangeVolume = createPropSetter("volume");
    this.onChangeDistanceModel = createPropSetter("distanceModel");
    this.onChangeRolloffFactor = createPropSetter("rolloffFactor");
    this.onChangeRefDistance = createPropSetter("refDistance");
    this.onChangeMaxDistance = createPropSetter("maxDistance");
    this.onChangeConeInnerAngle = createPropSetter("coneInnerAngle");
    this.onChangeConeOuterAngle = createPropSetter("coneOuterAngle");
    this.onChangeConeOuterGain = createPropSetter("coneOuterGain");
  }
```

PR #390, opened 2018-11-29
Bump max upload size to 150mb

robertlong:


robertlong:
Looks like this is merged in Reticulum now so I'll merge it.

PR #423, opened 2019-01-16
Switch from names to uuids for identifying nodes

### LINKED ISSUE #389: weird naming issue
I imported a model and named it 'cake'. If I leave the cursor in the name field and press the enter key repeatedly, the name of the model in the hierarchy view adds a constantly incrementing number.

firefox nightlyscreensnapz102

### LINKED ISSUE #420: Adding more than 10 lights causes parts of scene to disappear
Not counting the Ambient light, adding more than 10 of any type of light to your scene causes items to disappear from view. In my case, it would make the main model I had first imported go invisible-- it's still listed in the hierarchy, but it does not draw nor export properly.
This appears to be related to the problems with objects being auto-named despite my attempts to name them manually.
Example:  
1. Create Light1
2. Duplicate Light1:  Light2
3. Duplicate Light2: Light3
4. Try to rename Light2 to 'Light0':   It becomes Light4 
etc...

robertlong:
This is kinda a scary but long awaited change. This PR removes the requirement for unique node names by switching to UUIDs to identify nodes.

- Spoke scenes are now versioned.
  - Version 2 of the spoke scene uses uuids instead of names as the keys in the `entities` object and `root` property. It requires all entities to have a `name` property that specifies the node's name.
  - Version 1 scenes will be migrated to version 2 scenes upon opening them.
  - Spoke will now save version 2 scenes which are not backwards compatible with previous versions of Spoke. This is due to the inclusion of the scene node in the entities object.
- Unique names are no longer necessary inside the editor. You can have duplicate node names in a scene.
  - When duplicating a node, the new node will have a unique number appended to it. This is done for UX, so that you can differentiate between nodes that you have duplicated.
- When editing a node name and clicking on another node in the hierarchy view I noticed the edit was applied to the clicked node. This is due to the mouse down event firing before the blur event, so I changed it to use `onMouseUp`.

Closes #389
Closes #420


gfodor:
nice!

src/client/editor/utils/sortEntities.js
```diff
@@ -1,30 +1,30 @@
 export default function sortEntities(entitiesObj) {
   // Sort entities by insertion order
-  const sortedEntityNames = [];
```
robertlong: Changes to this file are just renaming things from name -> id to better reflect the data structure.

src/client/ui/hierarchy/HierarchyPanelContainer.js
```diff
@@ -114,7 +114,7 @@ class HierarchyPanelContainer extends Component {
           className: classNames("node", {
             "is-active": this.props.editor.selected && node.object.id === this.props.editor.selected.id
           }),
-          onMouseDown: e => this.onMouseDownNode(e, node)
+          onMouseUp: e => this.onMouseUpNode(e, node)
```
robertlong: This is the extra bug fix I snuck in and referred to in the PR. It directly affects renaming nodes and it's small enough that I don't feel like it needs to go in another PR 🤷‍♂️

PR #429, opened 2019-01-29
Navmesh Revamp

### LINKED ISSUE #274: Nav mesh should allow regeneration and settings once behavior is in scene

### LINKED ISSUE #307: Object Visiblility
There should be a way to temporarily hide an object. I added a new object and I wasn't sure where it was. I wanted to hide the landscape I already added so I could see it. But there wasn't a way.  There should be a visibility toggle like in photoshop layers.

robertlong:
- Added the visible checkbox to every node (except the scene)
- Added the walkable checkbox to model, box collider, and ground plane nodes.
- Added the collidable checkbox to model nodes.
- A floor plan node is included in the default new scene.
- Floor plan nodes can be configured and regenerated in the properties pane.
- Old scenes will have their nodes migrated to include walkable, collidable, visible, and floor-plan components accordingly.
- The floorplan's walkable area has been reduced so that it can be included in the nav mesh generation.
- `serialize()` and `prepareForExport()` were refactored to dry up adding the `visible` component when necessary.

Fixes #307 
Fixes #274

### DIFF
--- a/src/client/ui/App.js
+++ b/src/client/ui/App.js
@@ -201,6 +201,7 @@ class App extends Component {
   };
 
   componentDidMount() {
+    this.props.editor.project.watch();
     this.props.editor.signals.windowResize.dispatch();
     this.props.editor.signals.sceneModified.add(this.onSceneModified);
     this.props.editor.signals.editorError.add(this.onEditorError);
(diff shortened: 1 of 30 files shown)


src/client/ui/App.js
```diff
@@ -201,6 +201,7 @@ class App extends Component {
   };
 
   componentDidMount() {
+    this.props.editor.project.watch();
```
robertlong: Note: This line was removed in a previous PR and caused scenes to not upload properly.

PR #496, opened 2019-04-17
Add scene remixing (WIP)

robertlong:


emclaren:
Just noting that we had a request from the community to add an "edit" button in Hubs that would open up the scene in Spoke to edit/remix

PR #632, opened 2019-05-11
Add mouse look to fly controls

### LINKED ISSUE #616: Right click to enter fly mode when facing downwards after orbit mode jumps camera up <n/t>

johnshaughnessy:
Fixes #616


src/editor/controls/FlyControls.js
```diff
@@ -1,57 +1,104 @@
...
 
-    let xRotation = rotation.x > Math.PI ? rotation.x - 2 * Math.PI : rotation.x;
-    xRotation = THREE.Math.clamp(xRotation, -this.maxXRotation, this.maxXRotation);
+    // check change of local "forward" and "up" to disallow flipping
+    const camUpY = vec.set(0, 1, 0).applyQuaternion(worldQuat).y;
+    const newCamUpY = vec.set(0, 1, 0).applyQuaternion(candidateWorldQuat).y;
+    const newCamForwardY = vec.set(0, 0, -1).applyQuaternion(candidateWorldQuat).y;
+    const extrema = Math.sin(this.maxXRotation);
+    const allowRotationInX =
+      newCamUpY > 0 && ((newCamForwardY < extrema && newCamForwardY > -extrema) || newCamUpY > camUpY);
```
johnshaughnessy: If we only want to disallow flipping, use
`const allowRotationInX = newCamUpY > 0`
This instead clamps the rotation when it can, or if it is out of the clamp range (e.g. orbit controls left you looking straight down) it only allows rotation back towards the clamped range.

src/editor/controls/FlyControls.js
```diff
@@ -1,57 +1,104 @@
...
+      this.camera.matrixWorld.decompose(this.camera.position, this.camera.quaternion, this.camera.scale);
+    }
 
-    rotation.set(
-      xRotation + input.get(Fly.lookY) * lookSensitivity,
-      rotation.y + input.get(Fly.lookX) * lookSensitivity,
-      0,
-      "ZYX"
+    this.camera.matrixWorld.decompose(worldPos, worldQuat, worldScale);
+    // rotate about the world y axis
```
johnshaughnessy: this comment is a little misleading in that we aren't rotating the whole camera transform about the UP vector because we don't want it to translate. We are rotating it "in place" as if we translated the camera to the origin, rotated, then translated back.

robertlong:
LGTM 👍 thank you!

Issue #715, opened 2019-08-01
Projects should appear above Templates


brianpeiris:
It's annoying to have to scroll past Templates to view my projects. I suggest we move Templates below Projects, if you already have saved projects.

PR #731, opened 2019-08-20
Moving Projects section above templates

fcor:
Just moving projects section above Templates :)

Fixes https://github.com/mozilla/Spoke/issues/715


robertlong:
Thanks @fcor! What do you think about making the New Project button link to the templates page? I don't want the templates to get lost. I can add some new templates as well. Including a truly blank scene.

fcor:
Nice catch @robertlong! I was actually thinking something like that, for example some kind of prompt to select if starting from scratch or from a template. A blank template would be great too :).

Let me know if I can help with something else.

robertlong:
Thanks @fcor !

PR #750, opened 2019-09-11
Add stylelint and use styled-components v5 beta

robertlong:


robertlong:
It looks like v5 is broken in production builds. I'm keeping stylelint and the other changes but reverting bact to styled-components 4.3.2

PR #788, opened 2019-10-15
Feature/dynamic build configs

gfodor:
This updates spoke to support dynamic configuration for the build-time configs via meta tags. It also lifts HUBS_SERVER up into a proper configurable.


robertlong:
LGTM, is this ready to go now?

PR #801, opened 2019-10-19
Add tutorial to templates as "Open Terrain"

misslivirose:
Adds the tutorial project as an option named 'Open Terrain' to the projects page as a template.


robertlong:
Awesome, thanks!

PR #804, opened 2019-10-20
Properly handle modification state

gfodor:
This fixes some UX issues with scene modification state, one of which is arguably a P0 issue. Most of this behavior has regressed since we no longer were setting the `sceneModified` flag to true nor firing the `sceneModified` event. 

- The `onSceneModified` event now properly fires, which wasn't working before.

- We now properly save the project on publish if it is modified. Prior to this PR, if you edited a scene, published it, and reloaded, your changes would have been lost unless you hit Save.

- The modification state is now reflected in the title of the page

- We no longer nag you about leaving the page if you have not modified the scene

Also, I wasn't sure if there was some reason this behavior was specifically altered to be more conservative (wrt warning the user) -- this PR assumed that `objectsChanged` is a sufficient event to subscribe to to know if there has been a scene modification. lmk if that isn't right.

### DIFF
--- a/src/editor/Editor.js
+++ b/src/editor/Editor.js
@@ -205,8 +205,13 @@ export default class Editor extends EventEmitter {
 
     this.initialized = true;
 
-    this.addListener("objectsChanged", () => (this.sceneModified = true));
+    this.addListener("objectsChanged", this.onObjectsChanged);
     this.emit("initialized");
+  }
+
+  onObjectsChanged() {
+    this.sceneModified = true;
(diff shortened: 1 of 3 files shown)


robertlong:
This looks good, I added a listener for `sceneGraphChanged` which will ensure `sceneModified` gets set when reparenting objects.

This behavior was made more conservative due to complaints from users. However, it was working correctly at that point. After the refactor to support multi-select I believe this behavior regressed.

Thanks for the PR!

Issue #812, opened 2019-10-24
Add "Save as" functionality to create copies of scenes


emclaren:
It would be nice to be able to take a scene I already have, save a new version and modify it so the original stays untouched. Currently, users can export and import a spoke file, but it would be nice if there was an easier way.

Either a "Save as" button inside the project, or maybe a "Duplicate" button in the project thumbnail dropdown (see screenshot).

<img width="384" alt="Screen Shot 2019-10-24 at 11 22 01 AM" src="https://user-images.githubusercontent.com/6835324/67513870-963e6980-f650-11e9-9806-fdb2e00b450f.png">

tharun208:
Hi @misslivirose, If the button is as a thumbnail dropdown, we don't have much context about the project ( e.g current editor screen in-state). So, can we keep that inside the project? Looking forward to your comments on this.

robertlong:
@tharun208 Actually I think we can do this in the ProjectsPage in addition to the Save As option that you added to the editor menu.

Take a look at the `createProject` function in `Api.js` https://github.com/mozilla/Spoke/blob/master/src/api/Api.js#L418

 We can load the project file using the url returned from the projects API: `https://dev.reticulum.io/api/v1/projects` and then create a new project using that JSON and whatever new name we want to give it.

tharun208:
sure @robertlong Will Add these changes and raise a PR

tharun208:
But @robertlong, If we use the above API means we need to have the thumbnail image of the scene because it is required by the `createProject` API

robertlong:
@tharun208 yeah that is also returned by the `https://dev.reticulum.io/api/v1/projects` API. Rather than reusing the same thumbnail, I'd make a copy by downloading the file and reuploading. We should probably just have a server endpoint that does this, but until then we could implement it this way.

Here's an example response from that API endpoint:
```
{
  "projects": [
    {
      "name": "Tutorial",
      "project_id": "example",
      "project_url": "https://hubs-upload-cdn.com/files/example.json",
      "thumbnail_url": "https://hubs-upload-cdn.com/files/example.jpg"
    }
  ]
}
```

tharun208:
So right now we will be copying the same `project_url` and `thumbnail_url` and reusing it? @robertlong

PR #827, opened 2019-10-29
Added option for save as in editorContainer screen

tharun208:
This fixes issue #812. Added Save As Future.


src/ui/assets/MediaSourcePanel.js
```diff
@@ -26,7 +26,7 @@ export default function MediaSourcePanel({ editor, source, searchPlaceholder, in
 
       editor.spawnGrabbedObject(node);
     },
-    [editor]
+    [editor, source.transformPivot]
```
robertlong: Thanks for catching this! In the future, please submit fixes like this as a separate PR.

PR #833, opened 2019-10-30
Add Google Analytics Telemetry

robertlong:
I've added Google Analytics to Spoke so that we can track page views and our onboarding funnel conversion. I used the same techniques that we do in hubs to respect users privacy. We use the user's do not track settings to avoid loading the tracker in the first place and avoid sending any private data including the project name and id.


src/index.html
```diff
@@ -8,6 +8,17 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link href="https://fonts.googleapis.com/css?family=Lato:300,400,700" rel="stylesheet">
     <link href="https://fonts.googleapis.com/css?family=Zilla+Slab:400,700" rel="stylesheet">
+
+    <!-- Google Analytics -->
+    <!-- NOTE GA tracking is disabled by default. You'll need to set the GA_TRACKING_ID build variable to enable it. -->
+    <script>
+      if (navigator.doNotTrack !== "1") {
```
gfodor: if you haven't already make sure this matches exactly with the one in hubs, or else we have to update CSP

src/ui/App.js
```diff
@@ -69,15 +70,18 @@ export default class App extends Component {
...
-                  <Route path="/projects" exact component={ProjectsPage} />
-                  <Route path="/projects/:projectId" component={ProjectPage} />
-                  <Route path="/kits/package" component={PackageKitPage} />
+                  <Route path="/login" exact component={withPageView(LoginPage)} />
+                  <Route path="/logout" exact component={withPageView(LogoutPage)} />
+                  <Route path="/projects/templates" exact component={withPageView(TemplatesPage)} />
+                  <Route path="/projects" exact component={withPageView(ProjectsPage)} />
+                  <Route
+                    path="/projects/:projectId"
+                    component={withPageView(ProjectPage, "/projects/editor", "Editor")}
```
gfodor: +1

PR #834, opened 2019-10-31
Add Architecture Kit Example Template

robertlong:
I still need to look into project load performance before we can merge this.


robertlong:
Closing this since we've switched from templates to scene remixing!

PR #840, opened 2019-11-08
Use normalize-wheel package to fix scroll speed

### LINKED ISSUE #839: Mouse scrolling in the viewport is too extreme
I haven't done anything specific to my mouse settings (and the scrolling is fine / normal in other applications) but now when I use the mouse wheel in Spoke, the view port scrolling is too extreme. 

Video shows what two scroll 'notches' results in when the ground mesh is selected from the 'Wide Open Space' template, but this also happens when selecting spawn points. 

https://vimeo.com/371921522

Firefox Stable 70.0.1, Windows

robertlong:
Fixes #839

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "jwt-decode": "^2.2.0",
     "markdown-it": "^9.1.0",
     "mousetrap": "^1.6.3",
+    "normalize-wheel": "^1.0.1",
     "phoenix": "^1.4.10",
     "prop-types": "^15.7.2",
     "rc-slider": "^8.7.1",
--- a/src/editor/controls/InputManager.js
+++ b/src/editor/controls/InputManager.js
@@ -1,5 +1,6 @@
 import Mousetrap from "mousetrap";
 import isInputSelected from "../utils/isInputSelected";
+import normalizeWheel from "normalize-wheel";
 
 const _globalCallbacks = {};
 const _originalStopCallback = Mousetrap.prototype.stopCallback;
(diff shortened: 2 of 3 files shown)


misslivirose:
Testing now - had to update some stuff on my laptop

misslivirose:
What I see now is that it zooms out (not quite as far, but still out) and then zooms back in to a closer distance

misslivirose:
Hm, yeah, after testing more, I think that the experience of it auto-zooming is less preferable than the original issue, although both of them are quite difficult to work around. I've never had this issue with this mouse before trying it on this computer, and I've checked the scroll settings to see if anything seemed off but other pages are okay.

robertlong:
Ok, sounds like there's some work to be done. Let's test more when you're back.

misslivirose:
I suspect that this is something to do with this particular mouse driver - it's started acting up in other applications too on this PC, but I don't see this mouse issue in Spoke on other PCs / with other mice.

robertlong:
I think @InfiniteLee saw it with Windows smooth scrolling turned on. Maybe try enabling / disabling that?

misslivirose:
The mouse settings seemed the same on both computers that I tested - I can try to dig a little more, but so far no leads.

misslivirose:
I had to download and turn off smooth scrolling directly through the mouse software (changing the Windows settings and Firefox settings didn't work, but turning it off through the mouse customization app did)

Logitech M185 mouse. One I turned smooth scrolling off through the Logitech software, production Spoke was fine

robertlong:
I'm going to close this PR. If we have more issues with scrolling we can look into it again.

PR #847, opened 2019-11-19
Feature/spoke debranding

gfodor:
For self hosted instances, we can't have any Mozilla/Hubs/Spoke marks or branding.

Adds a build-time config `IS_MOZ` that flips spoke into one of two modes depending on if we should show Spoke/Hubs/Mozilla branding or not. It also turns off a few things like links to our Discord if in the non-branded mode. I made a few editorial decisions here, but don't have strong opinions about them so am happy to change as desired.

Un-branded mode is now the default. To run as branded spoke, do `yarn start-moz`


src/configs.js
```diff
@@ -16,10 +19,17 @@ get(configs, "CORS_PROXY_SERVER", process.env.CORS_PROXY_SERVER);
 get(configs, "NON_CORS_PROXY_DOMAINS", process.env.NON_CORS_PROXY_DOMAINS);
 get(configs, "SENTRY_DSN", process.env.SENTRY_DSN);
 get(configs, "BASE_ASSETS_PATH", process.env.BASE_ASSETS_PATH);
+get(configs, "IS_MOZ", process.env.IS_MOZ);
 
+console.log(process.env.IS_MOZ);
```
robertlong: You missed a console.log

src/ui/SpokeLogo.js
```diff
@@ -1,38 +1,50 @@
...
-        <path d="M101.686 119.09H48.324L75 149.445L101.686 119.09Z" fill="#0053BF" />
-        <path d="M1.16026 51.5853L27.1463 60.0104L31.3589 21.4912L3.31357 48.7003L1.16026 51.5853Z" fill="#66A8FF" />
-        <path d="M31.223 17.9684L28.7456 14.498L13.0244 35.6339L31.223 17.9684Z" fill="#66A8FF" />
-        <path d="M7.43204 99.3768L44.2892 116.112L27.6899 64.7461L7.43204 99.3768Z" fill="#003780" />
-        <path d="M0 58.2438L6.67944 96.5121L25.9965 63.4808L0 55.0557V58.2438Z" fill="#003780" />
-        <path d="M0 100.683L4.50523 99.2191L0 73.4316V100.683Z" fill="#003780" />
-        <path d="M5.16378 101.206L0.669006 102.659L17.0906 124.746L5.16378 101.206Z" fill="#003780" />
-      </svg>
-    );
+    if (configs.isMoz()) {
```
robertlong: This should probably just load the svg via a url rather than embedding it. Otherwise we're still loading the Spoke logo in non-moz contexts and it's not configurable.

src/ui/navigation/Footer.js
```diff
@@ -70,31 +71,41 @@ export default class Footer extends Component {
...
+              <NavListItem>
+                <a href="https://github.com/mozilla/hubs/blob/master/TERMS.md" rel="noopener noreferrer">
+                  Terms of Use
+                </a>
+              </NavListItem>
+            )}
+            {configs.isMoz() && (
+              <NavListItem>
+                <a href="https://github.com/mozilla/hubs/blob/master/PRIVACY.md" rel="noopener noreferrer">
+                  Privacy Notice
```
robertlong: Should these urls be configurable?

src/ui/SpokeLogo.js
gfodor: Yeah, I am leaving user-configurability of Spoke assets/links/etc to a second pass. This was just meant to get us to MVP of being able to ship an un-branded version so we can launch. I didn't tool out the plumbing to read the application configs as we do in hubs.

src/ui/navigation/Footer.js
gfodor: see above, we will add user-configuration of spoke to a later update.

gfodor:
I think the default images + videos are fine for now

PR #863, opened 2020-01-01
Fix attribution when republishing

robertlong:
Currently the creator attribution is pulled from a deprecated key on the scene. This PR pulls from the correct creator attribution when it exists.

Creator attribution should then:
- Use the existing value when republishing
- Use the value saved in local storage when publishing a new scene
- For legacy scenes without a scene associated with their project (projects previously stored the scene id in the project file itself, not in the db), the value stored in the project file's metadata will be used
- Otherwise be an empty string, including when remixing an existing scene because the creator attribution is held in the parent scene


robertlong:
Wrong base branch. Closing.

PR #865, opened 2020-01-02
Scene Remixing

robertlong:
You can now remix Hubs scenes created by the Hubs team and members of the Hubs community from the new project page! To get your scene featured in the new project page, check the "allow remixing" and "allow promotion" checkboxes when publishing your scene.

SceneRemixing


src/ui/router/ScrollToTop.js
```diff
@@ -0,0 +1,12 @@
+import { useEffect } from "react";
+import { useLocation } from "react-router-dom";
+
+export default function ScrollToTop() {
```
netpro2k: Already asked above, but why is this treated as a non-rendering component instead of a "custom hook"

src/ui/projects/CreateProjectPage.js
```diff
@@ -0,0 +1,164 @@
...
+    },
+    [history]
+  );
+
+  const { loading, error, entries, hasMore, loadMore } = usePaginatedSearch(
+    `${api.apiURL}/api/v1/media/search`,
+    params
+  );
+
+  const filteredEntries = entries.map(result => ({
```
netpro2k: Based on the name I was expecting this to filter out some results. Looks like its just cleaning up the response a bit. Might make sense to do this upstream somewhere, maybe in api.js (though as written `usePaginatedSearch` doesn't end up going through api.js so that would need to be reworked also).

src/ui/projects/CreateProjectPage.js
```diff
@@ -0,0 +1,164 @@
...
+                  <InfiniteScroll
+                    initialLoad={false}
+                    pageStart={0}
+                    loadMore={loadMore}
+                    hasMore={hasMore}
+                    threshold={100}
+                    useWindow={true}
+                  >
+                    <ProjectGrid
+                      projects={filteredEntries}
```
netpro2k: Bit weird that these are actually scene listings and not projects. Kind of surprising that it happens to work out. Would be pretty easy to mess up in the future though if their shapes diverge more.

src/ui/projects/CreateProjectPage.js
```diff
@@ -0,0 +1,164 @@
...
+      }
+
+      history.push(`/projects/create?${search}`);
+
+      setParams(nextParams);
+    },
+    [history]
+  );
+
+  const onChangeQuery = useCallback(
```
netpro2k: Seems weird you can only text search "all", but in checking it seems this is in line with how other media searches work...

src/ui/projects/CreateProjectPage.js
```diff
@@ -0,0 +1,164 @@
...
+                  <SearchInput placeholder="Search scenes..." value={params.q} onChange={onChangeQuery} />
+                </ProjectGridHeaderRow>
+                <ProjectGridHeaderRow>
+                  <Button as={Link} to="/projects/new">
+                    New Empty Project
+                  </Button>
+                </ProjectGridHeaderRow>
+              </ProjectGridHeader>
+              <ProjectGridContent>
+                <ScrollToTop />
```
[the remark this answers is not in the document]
robertlong: Yeah, maybe there's a hook for that, but this component comes straight from the React Router docs. It's pretty clever. https://reacttraining.com/react-router/web/guides/scroll-restoration

src/ui/projects/NewProjectGridItem.js
```diff
@@ -4,18 +4,20 @@ import PropTypes from "prop-types";
 import styled from "styled-components";
 import { Plus } from "styled-icons/fa-solid/Plus";
 
-const StyledNewProjectGridItem = styled(Link)`
+const ProjectGridItemContainer = styled.div`
```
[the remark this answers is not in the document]
robertlong: Yeah, I'm pretty bad at naming these random styled divs. I'll see if I can come up with a better name.

src/ui/projects/CreateProjectPage.js
robertlong: The problem is that projects and the media search API return different shaped responses. So this just maps scenes into something shaped like a project. Once we have a third type of media that uses this grid (I think that'll come soon) I'll refactor this and make the grid more generic.

src/ui/projects/CreateProjectPage.js
robertlong: Yeah, agreed, see above. I'll refactor with the next type of media to use this grid component.

src/ui/projects/CreateProjectPage.js
robertlong: Yeah, I figured that was the UX we wanted since we do it elsewhere. I'm not sure you usually want to search featured remixable scenes. Most of the time you want to search all remixable scenes.

PR #875, opened 2020-01-17
Water and Fog

robertlong:
You can now add water and fog to your Spoke scenes!

WaterAndFog

Adds the simple-water, fog, and background components.

Simple Water uses a modified MeshPhysicalMaterial on high graphics settings and MeshPhongMaterial on low. It has dynamic mesh deformation using simplex noise and two octaves for controlling swell and tide. It has specularity on both high and low settings but only receives shadows on high. It doesn't have reflections or refraction due to performance concerns, but both could be added in the future.

There are two fog types: linear and exponential. These can be used to limit the view distance in your scene or provide a nice atmosphere.

The background color can now be adjusted. When you have no skybox it can be useful to set a different background color. When there is no skybox and fog is enabled, this background color should usually be the same as the fog color.


src/editor/objects/SimpleWater.js
```diff
@@ -0,0 +1,250 @@
+import { Mesh, PlaneBufferGeometry, MeshStandardMaterial, MeshPhongMaterial, Vector2, RepeatWrapping } from "three";
```
netpro2k: If this file is intended to remain sync 1:1 with the one in hubs (which it looks like it is) we should add a comment here making note of that. It could also be moved to lib-hubs but that's obviously more work.

robertlong: Yeah I want to move `SimpleWater` and `Sky` to `lib-hubs` I'll add the comment for now.

BluePrintRandom:
\0/

PR #887, opened 2020-02-07
Experimental Preview Mode

robertlong:
You can now preview model animations, videos, and audio in Spoke! To enable this experimental feature go to File -> Developer -> Enable Experimental Features. You'll then see a play button in the toolbar that you can use to play/pause the preview.

preview-mode

### DIFF
--- a/src/editor/Editor.js
+++ b/src/editor/Editor.js
@@ -562,6 +562,7 @@ export default class Editor extends EventEmitter {
         node.onPlay();
       }
     });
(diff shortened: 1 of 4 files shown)


blairmacintyre:
I tried it on this scene, with 3 twitch streams, and while I see a single frame of the stream that is active right now, hitting play doesn't actually play the video.  I think it did _something_ one time (I briefly heard audio) but not most times.

PR #892, opened 2020-02-10
Fix bug with animations

robertlong:


gfodor:
lgtm, nice job cleaning up the test framework to cover it

gfodor:
looks like build is failing fyi

robertlong:
Yup, I got bit by another version of three getting linked into my project and causing `propValue instanceof Color` to fail 😞

PR #893, opened 2020-02-10
Scene Audio Settings

robertlong:
You can now adjust the audio settings for avatars and media in your scene. Select the scene node in the hierarchy panel and check "Override Audio Settings" to adjust these properties.

Screen Shot 2020-02-12 at 11 16 11 AM


.env.defaults
```diff
@@ -7,10 +7,10 @@
 # The Hubs backend to use for previewing published scenes.
 # The RETICULUM_SERVER must also be set to dev to avoid the CSP rules.
 # Uncomment both lines when testing with a local hubs instance.
-# HUBS_SERVER="localhost:8080"
+HUBS_SERVER="localhost:8080"
```
gfodor: intentional?

src/ui/properties/SceneNodeEditor.js
```diff
@@ -23,90 +27,245 @@ const FogTypeOptions = [
...
+          name="Fog Density"
+          smallStep={0.01}
+          mediumStep={0.1}
+          largeStep={0.25}
+          min={0}
+          value={node.fogDensity}
+          onChange={onChangeFogDensity}
+        />
+      )}
+      <InputGroup name="Override Audio Settings">
```
gfodor: Perhaps add `info` with a link to the pannernode settings page on MDN?

robertlong: Info is a tooltip, so you can't actually click on links in it. Changing the UX of prop info seems out of scope for this PR. However, I agree that would be a nice thing to add.

.env.defaults
robertlong: Nope, thanks!

PR #927, opened 2020-03-05
Add controls checkbox to image node. Set proper defaults.

robertlong:


src/editor/nodes/VideoNode.js
```diff
@@ -62,6 +62,7 @@ export default class VideoNode extends EditorNodeMixin(Video) {
     this._canonicalUrl = "";
     this._autoPlay = true;
     this.volume = 0.5;
+    this.controls = false;
```
johnshaughnessy: Right now audio and video nodes will always have controls in hubs: https://github.com/mozilla/hubs/blob/master/src/gltf-component-mappings.js#L192

Not sure if we want to change this behavior. Personally I like being able to change the volume of videos/audio loaded from Spoke (and I like for them to be "focus-able", which right now is tied together), but if we want to have uncontrollable videos/audio nodes we'll have to edit hubs too.

src/editor/nodes/ImageNode.js
```diff
@@ -15,11 +15,12 @@ export default class ImageNode extends EditorNodeMixin(Image) {
   static async deserialize(editor, json, loadAsync, onError) {
     const node = await super.deserialize(editor, json);
 
-    const { src, projection } = json.components.find(c => c.name === "image").props;
+    const { src, projection, controls } = json.components.find(c => c.name === "image").props;
 
     loadAsync(
       (async () => {
         await node.load(src, onError);
+        node.controls = controls === undefined ? true : controls;
```
johnshaughnessy: Not sure which is a reasonable default : for images to have "controls" or not. For images that have been used so far as a way to texture a wall, it's not clear that the scene author wants menus to appear when you hover over the image in hubs.

robertlong: I'm going to change the default in the constructor to true and just coerce undefined to false here instead.

PR #946, opened 2020-04-13
Performance Tools

### LINKED ISSUE #549: Show current scene complexity
It is very sad to have a scene that you don't realize is too big, and then end up having to delete objects and work backwards. If we could recognize a few warning signs, it could make the experience a little more smooth for users who aren't familiar with some of the things they should do to optimize their scenes.

### LINKED ISSUE #696: Add glTF validator to Spoke and validate on importing models
We should run the glTF Validator on all imported models so we can better warn users that they will have an issue on export.

https://github.com/KhronosGroup/glTF-Validator

robertlong:
Creating scenes that run well on all devices has never been easier! It is now possible to diagnose performance issues before you publish using the Performance Check tool. This Spoke tool will allow you to see how resource-intensive your scenes are before you publish them. Items that are computationally heavy will also have warnings on them in the hierarchy panel so you can easily see which assets may cause performance impacts.

Screen Shot 2020-04-13 at 2 50 28 PM

 Check the viewport's current FPS and other renderer info using the stats overlay.

Screen Shot 2020-04-13 at 2 51 29 PM

Check file sizes, texture resolutions, and validate models in the glTF Info section. 

Screen Shot 2020-04-09 at 10 56 13 AM

And get realtime performance warnings in the Hierarchy Panel.

Screen Shot 2020-04-10 at 5 36 23 PM

Closes #696
Closes #549


misslivirose:
This is AMAZING and I'm so thrilled to see this land. Nice work! 

Copy suggestion for What's New: 



> Creating scenes that run well on all devices has never been easier! It is now possible to diagnose performance issues before you publish using the Performance Check tool. This Spoke tool will allow you to see how resource-intensive your scenes are before you publish them. Items that are computationally heavy will also have warnings on them in the hierarchy panel so you can easily see which assets may cause performance impacts.

robertlong:
Awesome, thanks! Updated.

PR #957, opened 2020-04-17
Modify the searchMedia() function in Api.js.

igorgark1593:


robertlong:
We're planning on adding blocklist support in Reticulum. For separate deployments we still looking into an extensibility API for Spoke to enable this.

robertlong:
This specific feature is covered in Hubs/Reticulum in this issue: https://github.com/mozilla/hubs/issues/2947

PR #968, opened 2020-05-23
Decouple Spoke from Hubs/Reticulum + add search term filter

shawticus:
Two separate ideas that got mixed up as a result of 4am code pushing -- increased decoupling from Reticulum and the Reticulum API structure (based on lessons we've learned integrating Spoke in a non-Hubs environment) and a media search filter (a client request we've had twice now, will argue that it's a very common use case below, although this might be a backend thing, not a Spoke thing... ).

Re: Decoupling

While most any sane backend engineer is going to use some Rails-y style MVC which follows some verb and endpoint conventions, it will take a lot of acrobatics to line up exactly with the Reticulum API if things are done different or if they're already using the routes another way and can't work around it.

This PR abstracts the route entirely to .env configuration so it can be configured to accommodate any reasonable API structure. There should be no noticeable change to Spoke.

Re: Blacklist

The idea here is that educational, corporate, and other PG content-oriented sectors will want a way to block NSFW content -- of which there is plenty on Sketchfab, etc.

This is a proposal -- it has minimal impact on the API, but the feature might make better sense implemented on the backend. The benefit here is that we can prevent the request from ever happening. Also it's a lot easier right now for the average user to ship and deploy custom Spoke > custom Reticulum into their AWS Marketplace instance, in my experience, so this may help out some of the people hopping on the Discord from time to time looking for this feature as a quick-fix with minimal impact.

The PR includes the terms 'sex' and 'drugs', but this could be set to null to default to no blacklist.

(A separate blacklist.txt was my original intention, but I'm not sure if we want to add the dependency on fs.)


avaer:
The decoupling part of this PR sounds promising.

In XRPackage we bundle Hubs as a deployable app (e.g. multi-Hubs, run Unity inside of Hubs etc). However this currently requires hacks to route around reticulum hard-ish dependencies.

https://github.com/webaverse/hubs/pull/1/files

Making this pluggable would make it much simpler for us to have a generic Hubs package that can connect to the different Hubs based on reticulum parameters. So I am conceptually for this.

robertlong:
I think we need to further discuss how to decouple API.js from Spoke and make it more configurable. The addition of a language filter in this PR isn't something we're looking to include in Spoke at this time and that'd be something you'd likely be able to configure inside API.js

Let's discuss this further in another PR.

PR #982, opened 2020-06-17
Fixes issue where Vanilla Spoke won't start

shawticus:
Issue: A fresh install of Spoke won't start
Cause: Corejs is out of date, and updated threejs no longer includes waternormals.jpg in npm package

Solution: included waternormals.jpg in assets and updated Corejs with this PR.


robertlong:
LGTM, thanks!

PR #987, opened 2020-07-03
Add "alpha mode" setting for images

netpro2k:
Adds the option to set transparency mode for images. By default new image elements will be set to "none". Any existing images will default to "alpha" to preserve the original behavior. Hubs is similarly configured to load media components without this setting defaulting to the old behavior (alpha).

image

Hubs PR https://github.com/mozilla/hubs/pull/2605

### DIFF
--- a/src/editor/nodes/ImageNode.js
+++ b/src/editor/nodes/ImageNode.js
@@ -24,8 +24,8 @@ export default class ImageNode extends EditorNodeMixin(Image) {
       (async () => {
         await node.load(src, onError);
         node.controls = controls || false;
-        node.transparencyMode = transparencyMode === undefined ? ImageTransparencyMode.Alpha : transparencyMode;
         node.alphaCutoff = alphaCutoff || 0.5;
+        node.transparencyMode = transparencyMode === undefined ? ImageTransparencyMode.Blend : transparencyMode;
         node.projection = projection;
       })()
     );
@@ -135,7 +135,7 @@ export default class ImageNode extends EditorNodeMixin(Image) {
       transparencyMode: this.transparencyMode,
       projection: this.projection
     };
-    if (this.transparencyMode === ImageTransparencyMode.Cutout) {
+    if (this.transparencyMode === ImageTransparencyMode.Mask) {
       imageData.alphaCutoff = this.alphaCutoff;
     }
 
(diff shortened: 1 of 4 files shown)


src/ui/layout/Tooltip.js
```diff
@@ -50,6 +50,8 @@ export const TooltipContainer = styled.div`
   overflow-wrap: break-word;
   user-select: none;
   text-align: center;
+  white-space: pre-wrap;
```
netpro2k, line 53: This is a quick way to allow formatted multi-line tooltips

src/editor/objects/Image.js
```diff
@@ -16,15 +15,25 @@ export const ImageProjection = {
   Equirectangular360: "360-equirectangular"
 };
 
+export const ImageTransparencyMode = {
+  None: "none",
+  Alpha: "alpha",
+  Cutout: "cutout"
```
robertlong, line 21: I'd rather we use the term alpha "mask" and "blend" as they correspond to the terminology already used in glTF: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#alpha-coverage

src/editor/nodes/ImageNode.js
```diff
@@ -16,12 +16,16 @@ export default class ImageNode extends EditorNodeMixin(Image) {
...
+    const { src, projection, controls, transparencyMode, alphaCutoff } = json.components.find(
+      c => c.name === "image"
+    ).props;
 
     loadAsync(
       (async () => {
         await node.load(src, onError);
         node.controls = controls || false;
+        node.transparencyMode = transparencyMode === undefined ? ImageTransparencyMode.Alpha : transparencyMode;
+        node.alphaCutoff = alphaCutoff || 0.5;
```
robertlong, line 28: You need to check for `undefined` here or else it'll deserialize 0 to 0.5.

src/editor/objects/Image.js
netpro2k: Makes sense, I changed it to `alphaMode` and changed the possible values to `opaque`, `blend`, and `mask` to align more with the gltf spec

robertlong:
Ah @netpro2k I forgot to have you update the test snapshots. This introduces some new properties to nodes in the scenes so the tests broke. For future reference you can update the test snapshots by running `yarn update-test-snapshots`

PR #989, opened 2020-07-07
Revert "Fixes issue where Vanilla Spoke won't start"

robertlong:
I should have read the PR more carefully. This is now breaking our builds so I'm going to roll this back.

I don't want us making all dependencies runtime dependencies and we use a yarn lockfile in this project. The waternormals fix looks valid though, I don't think that file is actually in the examples folder. `core-js` can be pinned to a newer version if necessary. Why does it need to be upgraded though?

@shawticus sorry for the 180, can you provide some more context so we can get this resolved?

Reverts mozilla/Spoke#982


shawticus:
My bad, this was a while ago but I believe this was just meant to be the
waternormals.jpg / reference change.

PR #998, opened 2020-07-22
Better describe what the Link element does.

SeanBannister:
I had no idea the Link element could link to a website because the description only mentioned it was for linking to hub rooms.


robertlong:
Thank you!

PR #1011, opened 2020-08-03
Add Scene Preview Camera Node

### LINKED ISSUE #284: Improve thumbnail screenshot UX
It's annoying to have to reposition yourself correctly every time you re-publish a scene. 
Probably the best UX is to allow you to position the camera correctly in the publish dialog (instead of a static thumbnail) and then reuse that camera position.

robertlong:
Spoke has always used the camera position when you save or publish to generate the thumbnail for a project and scene. This has been a major pain point for many users and now there's a solution. The Scene Preview Camera Node can be placed once in your scene and it will always be used to position the camera used for generating scene and project thumbnails. It also represents the starting position of the scene preview camera in Hubs. If you don't add a scene preview camera, one will be added for you the next time you publish your scene.

Fixes #284 
Closes https://github.com/mozilla/hubs/issues/2666

### DIFF
--- a/src/api/Api.js
+++ b/src/api/Api.js
@@ -557,7 +557,7 @@ export default class Project extends EventEmitter {
       throw new Error("Save project aborted");
     }
 
-    const { blob: thumbnailBlob } = await editor.takeScreenshot(512, 320);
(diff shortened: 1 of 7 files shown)


joshmarinacci:
Nice work

PR #1019, opened 2020-09-02
Media Frames

netpro2k:
Adds support for the "Media Frame" element, which is used to capture media and position it precisely in space. Requires the associated Hubs PR to go out first https://github.com/mozilla/hubs/pull/2967. Though the initial Hubs PR does not handle networking, we write out a network id for this component in anticipation we will be needing it in the future.

As with the Hubs side, the visuals for the helpers could use some love, so we will need to update them once we land on something better for Hubs.


src/editor/nodes/MediaFrameNode.js
```diff
@@ -0,0 +1,141 @@
...
+  copy(source, recursive = true) {
+    if (recursive) {
+      this.remove(this.helper);
+    }
+
+    this.mediaTypes = source.mediaTypes;
+
+    super.copy(source, recursive);
+
+    if (recursive) {
```
netpro2k, line 104: Cargo culted this from the rest of the nodes, but don't really get whats going on here.

robertlong: The `clone()` method calls the constructor which creates your helper object. The `copy()` method also clones the child elements if the recursive parameter is passed. It is passed when cloning the scene or duplicating an object in the editor. So we need to remove the helper object before copying and then re-associate the cloned helper object after the copy method is called. This looks correct :+1:

PR #1027, opened 2020-09-14
Update maps transforms if available in GLTF

### LINKED ISSUE #999: Spoke ignores KHR_texture_transform extension
When exporting an object, say, from Blender that uses the Mapping node to tile a texture, Spoke appears to ignore it and keep whatever the basic mapping coordinates it had.
Hubs, by contrast, keeps those coordinates when dropping the object in.

Examples:
Here is the object with no Mapping node added.
UVsNoTransform

Here is the object with a Mapping node added and set to tile 4 times in x & y:
UVsTransformed

After exporting that tiled object with Custom Properties checked in order to gain a KHR_texture_transform (as per the Blender documentation):
In Hubs (correct):
UVsTransformed_Hubs

In Spoke (incorrect):
UVsTransformed_Spoke

keianhzo:
Fixes #999 This PR updates the ThreeJS material maps tranforms based on the available maps transforms when a GLTF model is loaded.

### DIFF
--- a/src/editor/gltf/GLTFLoader.js
+++ b/src/editor/gltf/GLTFLoader.js
@@ -167,6 +167,24 @@ export const MIME_TYPE_FORMATS = {
 };
 
 /*********************************/
+/********** GLTF EXTENSIONS ***********/
+/*********************************/
+
+const EXTENSIONS = {
+  KHR_BINARY_GLTF: "KHR_binary_glTF",
+  KHR_DRACO_MESH_COMPRESSION: "KHR_draco_mesh_compression",
+  KHR_LIGHTS_PUNCTUAL: "KHR_lights_punctual",
+  KHR_MATERIALS_CLEARCOAT: "KHR_materials_clearcoat",
+  KHR_MATERIALS_PBR_SPECULAR_GLOSSINESS: "KHR_materials_pbrSpecularGlossiness",
+  KHR_MATERIALS_TRANSMISSION: "KHR_materials_transmission",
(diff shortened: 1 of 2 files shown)


src/editor/gltf/GLTFLoader.js
```diff
@@ -1446,9 +1446,41 @@ class GLTFLoader {
...
 
+  /**
+   * Applies a texture transform, if present, to the map definition. Requires
+   * the KHR_texture_transform extension.
+   */
+  applyTextureTransform(mapDef, map) {
+    if (!mapDef || !map || !mapDef.extensions || !mapDef.extensions["KHR_texture_transform"]) {
+      return;
+    }
+
```
robertlong, line 1466: In the Three.js implementation, they clone the texture object and set `texture.needsUpdate = true` I think we'll want to do the same just to be sure. https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/GLTFLoader.js#L812

robertlong (commented):
This looks pretty close! I think we'll just want to adopt a few of the things I mentioned from the Three.js implementation.

src/editor/gltf/GLTFLoader.js
keianhzo: I've updated to make it more similar to the ThreeJS loader.

robertlong:
LGTM 👍 I'll merge and deploy it.

keianhzo:
@robertlong I've updated to only combine meshes if transforms match.

robertlong:
Perfect! Great work!

PR #1028, opened 2020-09-15
Make sure the extensions object is present before adding the extension

### LINKED ISSUE #1023: Lightmaps get ignored on export when using PBR shader
If you have a model using a built-in lightmap (exported from Blender using our hubs-blender-exporter add-on), if you are using a BSDF shader (PBR), it will be ignored by Spoke when publishing to Hubs.
The lightmap will show as expected inside Hubs, so there must be something getting ignored or stripped out at publish time.
(Two .glb objects attached for reference)

The object on the right is using an unlit shader (Background node in Blender), whereas the lefthand object is using the Standard material (BSDF node in Blender):
image

Note: Hubs will recognize the object's lightmaps correctly if you simply drag/drop them into Hubs:

(Foreground is the dropped-in object):
image

[Lightmapped_Objects.zip](https://github.com/mozilla/Spoke/files/5198279/Lightmapped_Objects.zip)

keianhzo:
Fixes #1023 Make sure the extensions object is present before adding the extension

### DIFF
--- a/src/editor/gltf/extensions/exporter/LightmapExporterExtension.js
+++ b/src/editor/gltf/extensions/exporter/LightmapExporterExtension.js
@@ -7,6 +7,9 @@ export class LightmapExporterExtension extends ExporterExtension {
       () => true,
       (material, materialDef) => {
         if (material.lightMap) {
+          if (materialDef.extensions === undefined) {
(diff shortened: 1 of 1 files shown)


robertlong:
LGTM 👍

PR #1029, opened 2020-09-17
Support multiple animations on a single model

### LINKED ISSUE #924: Support multiple animations on a single model
It's pretty common in Blender to export an object with multiple animations and want more than one to run at a time. We should enable that in Spoke.

Ideally you would select multiple animations from the select input in a model's properties panel. This UI should be as simple as using react-select's [multi-select option](https://react-select.com/home).

This needs to be configured in the [ModelNodeEditor](https://github.com/mozilla/Spoke/blob/master/src/ui/properties/ModelNodeEditor.js#L65).

Then [Model](https://github.com/mozilla/Spoke/blob/master/src/editor/objects/Model.js) and [ModelNode](https://github.com/mozilla/Spoke/blob/master/src/editor/nodes/ModelNode.js) need to be setup to handle multiple animations. And finally, these extra animations need to be exported. This is handled by passing animation clips to the GLTFExporter [here](https://github.com/mozilla/Spoke/blob/master/src/editor/Editor.js#L396). We can collect the full set of used animations in the SceneNode [here](https://github.com/mozilla/Spoke/blob/master/src/editor/nodes/SceneNode.js#L575).

Then on the Hubs side we need to change how the [loop-animation](https://github.com/mozilla/hubs/blob/master/src/components/loop-animation.js#L49) component handles deserializing multiple animations.

keianhzo:
Spoke now supports multiple animations per model. After importing a model from Sketchfab or your favorite 3D tool, you can now select multiple animations to loop in the Spoke property panel.

There is also a new checkbox to ignore mesh combination for a specific model. This can be useful when you want to add dynamic behavior to an object in your custom Hubs Cloud client.

image

Fixes #924 Support multiple animations on a single model

Depends on https://github.com/mozilla/hubs/pull/3061

### DIFF
--- a/src/editor/nodes/EditorNodeMixin.js
+++ b/src/editor/nodes/EditorNodeMixin.js
@@ -208,34 +208,72 @@ export default function EditorNodeMixin(Object3DClass) {
       }
     }
 
-    addGLTFComponent(name, props) {
-      if (!this.userData.gltfExtensions) {
-        this.userData.gltfExtensions = {};
-      }
+    /**
+     * Adds a GLTF component
+     * @param {String} name The component name to be replaced
+     * @param {Object} props The component properties to be set
+     * @param {boolean} params Parameters object
+     * @param {{THREE.Object3D} params.node The target Object3D element
+     * @param {boolean} params.replace Wether or not the component should be replaced or just added if it does not exist
+     * @param {boolean} params.recursive Call this method recursively on all the Object3D children
+     */
+    addGLTFComponent(name, props, params = {}) {
+      params.node = params.node || this;
+      params.replace = params.replace || false;
+      params.recursive = params.recursive || false;
 
-      if (!this.userData.gltfExtensions.MOZ_hubs_components) {
-        this.userData.gltfExtensions.MOZ_hubs_components = {};
+      let update = false;
+      if (params.replace) {
+        if (
+          params.node.userData.gltfExtensions &&
+          params.node.userData.gltfExtensions.MOZ_hubs_components &&
+          params.node.userData.gltfExtensions.MOZ_hubs_components[name]
+        ) {
+          update = true;
+        }
+      } else {
+        if (!params.node.userData.gltfExtensions) {
+          params.node.userData.gltfExtensions = {};
+        }
+
+        if (!params.node.userData.gltfExtensions.MOZ_hubs_components) {
+          params.node.userData.gltfExtensions.MOZ_hubs_components = {};
+        }
+
+        update = true;
       }
 
       if (props !== undefined && typeof props !== "object") {
         throw new Error("glTF component props must be an object or undefined");
       }
 
-      const componentProps = {};
+      if (update) {
+        const componentProps = {};
 
-      for (const key in props) {
-        if (!Object.prototype.hasOwnProperty.call(props, key)) continue;
+        for (const key in props) {
+          if (!Object.prototype.hasOwnProperty.call(props, key)) continue;
 
-        const value = props[key];
+          const value = props[key];
 
-        if (value instanceof Color) {
-          componentProps[key] = serializeColor(value);
-        } else {
-          componentProps[key] = value;
+          if (value instanceof Color) {
+            componentProps[key] = serializeColor(value);
+          } else {
+            componentProps[key] = value;
+          }
         }
+
+        params.node.userData.gltfExtensions.MOZ_hubs_components[name] = componentProps;
       }
 
-      this.userData.gltfExtensions.MOZ_hubs_components[name] = componentProps;
+      if (params.recursive) {
+        params.node.children.forEach(child => {
+          this.addGLTFComponent(name, props, {
+            node: child,
+            replace: params.replace,
+            recursive: params.recursive
+          });
+        });
+      }
     }
 
     replaceObject(replacementObject) {
--- a/src/editor/nodes/EditorNodeMixin.js
+++ b/src/editor/nodes/EditorNodeMixin.js
@@ -214,66 +214,35 @@ export default function EditorNodeMixin(Object3DClass) {
      * @param {Object} props The component properties to be set
      * @param {boolean} params Parameters object
      * @param {{THREE.Object3D} params.node The target Object3D element
-     * @param {boolean} params.replace Wether or not the component should be replaced or just added if it does not exist
-     * @param {boolean} params.recursive Call this method recursively on all the Object3D children
      */
-    addGLTFComponent(name, props, params = {}) {
-      params.node = params.node || this;
-      params.replace = params.replace || false;
-      params.recursive = params.recursive || false;
+    addGLTFComponent(name, props) {
+      if (!this.userData.gltfExtensions) {
+        this.userData.gltfExtensions = {};
+      }
 
-      let update = false;
-      if (params.replace) {
-        if (
-          params.node.userData.gltfExtensions &&
-          params.node.userData.gltfExtensions.MOZ_hubs_components &&
-          params.node.userData.gltfExtensions.MOZ_hubs_components[name]
-        ) {
-          update = true;
-        }
-      } else {
-        if (!params.node.userData.gltfExtensions) {
-          params.node.userData.gltfExtensions = {};
-        }
-
-        if (!params.node.userData.gltfExtensions.MOZ_hubs_components) {
-          params.node.userData.gltfExtensions.MOZ_hubs_components = {};
-        }
-
-        update = true;
+      if (!this.userData.gltfExtensions.MOZ_hubs_components) {
+        this.userData.gltfExtensions.MOZ_hubs_components = {};
       }
 
       if (props !== undefined && typeof props !== "object") {
         throw new Error("glTF component props must be an object or undefined");
       }
 
-      if (update) {
-        const componentProps = {};
+      const componentProps = {};
 
-        for (const key in props) {
-          if (!Object.prototype.hasOwnProperty.call(props, key)) continue;
+      for (const key in props) {
+        if (!Object.prototype.hasOwnProperty.call(props, key)) continue;
 
-          const value = props[key];
+        const value = props[key];
 
-          if (value instanceof Color) {
-            componentProps[key] = serializeColor(value);
-          } else {
-            componentProps[key] = value;
-          }
+        if (value instanceof Color) {
+          componentProps[key] = serializeColor(value);
+        } else {
+          componentProps[key] = value;
         }
-
-        params.node.userData.gltfExtensions.MOZ_hubs_components[name] = componentProps;
       }
 
-      if (params.recursive) {
-        params.node.children.forEach(child => {
-          this.addGLTFComponent(name, props, {
-            node: child,
-            replace: params.replace,
-            recursive: params.recursive
-          });
-        });
-      }
+      this.userData.gltfExtensions.MOZ_hubs_components[name] = componentProps;
     }
 
     replaceObject(replacementObject) {
--- a/src/editor/nodes/KitPieceNode.js
+++ b/src/editor/nodes/KitPieceNode.js
@@ -28,6 +28,7 @@ export default class KitPieceNode extends EditorNodeMixin(Model) {
 
         node.collidable = !!json.components.find(c => c.name === "collidable");
         node.walkable = !!json.components.find(c => c.name === "walkable");
+        node.combine = !!json.components.find(c => c.name === "combine");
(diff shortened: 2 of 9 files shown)


keianhzo:
Simple model to help testing

[multiple_anim.zip](https://github.com/mozilla/Spoke/files/5238998/multiple_anim.zip)

src/editor/nodes/ModelNode.js
```diff
@@ -351,26 +357,23 @@ export default class ModelNode extends EditorNodeMixin(Model) {
...
-          `Error exporting model "${this.name}" with url "${this._canonicalUrl}". Animation could not be found.`
-        );
-      } else {
-        this.addGLTFComponent("loop-animation", {
-          activeClipIndex: activeClipIndex
-        });
+    this.addGLTFComponent(
+      "loop-animation",
+      {
+        clip: this.activeClipNames.join(),
```
robertlong, line 370: Rather than relying on the clip name, I'd rather we focus on exporting an array of clip indices and modify Hubs to read those indices.

So it'd look something like:
```js
if (this.activeClips.length > 0) {
  const activeClipIndices = this.activeClips.map((clip) => {
    const index = ctx.animations.indexOf(clip);

    if (index === -1) {
      throw new Error(`Error exporting model "${this.name}" with url "${this._canonicalUrl}". Animation could not be found.`);
    }

    return index;
  });


  this.addGLTFComponent("loop-animation", { activeClipIndices });
}
```

robertlong (commented):
I think we should keep storing animations not by the animation name, but by the animation index. People bring in models from all sorts of sources and the animations aren't guaranteed to be unique. So storing animations by index lets us ensure that no matter what asset the user brings in, it'll work.

As we move from a single animation to multiple animations, we'll want to switch from storing a single integer to an array of integers representing each animation index. We can call this field `activeClipIndices `, which states: these are the animation clips that should play when the scene is loaded, or these clips are active.

This change will also require a change to Hubs. Right now we only support a single integer for animation index, but we can move to an array of indices: https://github.com/mozilla/hubs/blob/master/src/components/loop-animation.js#L12

The exported gltf properties will be passed through to the Aframe `loop-animation` component. So all that is needed to make use of this array is to add a property to the aframe schema and to use the new property. You can define the new `activeClipIndices` property with the `array` property type: https://aframe.io/docs/1.0.0/core/component.html#schema_array

Let me know if you want to pair on this. I can make myself available at a reasonably early hour PST if that works best for you.

src/editor/nodes/EditorNodeMixin.js
```diff
@@ -208,13 +208,38 @@ export default function EditorNodeMixin(Object3DClass) {
...
-        this.userData.gltfExtensions = {};
-      }
+    /**
+     * Adds a GLTF component
+     * @param {String} name The component name to be replaced
+     * @param {Object} props The component properties to be set
+     * @param {boolean} params Parameters object
+     * @param {{THREE.Object3D} params.node The target Object3D element
+     * @param {boolean} params.replace Wether or not the component should be replaced or just added if it does not exist
+     * @param {boolean} params.recursive Call this method recursively on all the Object3D children
```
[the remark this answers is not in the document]
keianhzo: We need to make sure that only the node that originally had the loop-animation component is exported with that component. We were currently adding the loop-animation node to the top level model node but in the original GLTF file the child armature node is the one with the loop-animation node so we are not updating the right node upon export and the original GLTF animations are played in Hubs.

There was a bug in the code and the component was being added even if it didn't exist, I've fixed that.

src/editor/nodes/ModelNode.js
keianhzo: Updated

src/editor/nodes/ModelNode.js
```diff
@@ -315,11 +315,17 @@ export default class ModelNode extends EditorNodeMixin(Model) {
       }
     };
 
-    if (this.activeClipIndex !== -1) {
-      components["loop-animation"] = {
-        activeClipIndex: this.activeClipIndex
-      };
-    }
+    this.addGLTFComponent(
```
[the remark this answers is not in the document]
keianhzo: Fixed.

keianhzo:
@robertlong I've updated and now this requires https://github.com/mozilla/hubs/pull/3061

src/editor/nodes/EditorNodeMixin.js
```diff
@@ -208,34 +208,73 @@ export default function EditorNodeMixin(Object3DClass) {
...
+     * @param {boolean} params.firstPass If this is the first execution of the method
+     */
+    addGLTFComponent(name, props, params = {}) {
```
robertlong, line 221: @keianhzo can you explain the need for adding components recursively? I'm not sure what this is doing.

src/editor/nodes/KitPieceNode.js
```diff
@@ -38,7 +38,7 @@ export default class KitPieceNode extends EditorNodeMixin(Model) {
             node.activeClipIndex = loopAnimationComponent.props.activeClipIndex;
           } else if (clip !== undefined && node.model && node.model.animations) {
             // DEPRECATED: Old loop-animation component stored the clip name rather than the clip index
-            node.activeClipIndex = node.model.animations.findIndex(animation => animation.name === clip);
+            node.activeClipIndex = [node.model.animations.findIndex(animation => animation.name === clip)];
```
robertlong, line 41: Can we change this to `activeClipIndices`?

src/editor/nodes/ModelNode.js
```diff
@@ -351,26 +349,22 @@ export default class ModelNode extends EditorNodeMixin(Model) {
...
-          `Error exporting model "${this.name}" with url "${this._canonicalUrl}". Animation could not be found.`
-        );
-      } else {
-        this.addGLTFComponent("loop-animation", {
-          activeClipIndex: activeClipIndex
-        });
+    this.addGLTFComponent(
+      "loop-animation",
+      {
+        activeClipIndices: this.activeClipIndices
```
robertlong, line 362: I think we need to use `this.activeClips.map(clip => ctx.animations.indexOf(clip))` otherwise multiple animated objects will point to the same animation clip. The `ctx.animations` contains all of the animations in the exported glTF scene in the order it will be output.

robertlong (commented):
Sorry, just a few more things we should look into.

src/editor/nodes/EditorNodeMixin.js
keianhzo: The recursive flag is used to traverse the node hierarchy. When used along with the replace flag, you make sure you update the node holding a specific component instead of just adding it to the top level node as we were doing.

The loop-animation component has to be added to the Armature node in Blender to work in Hubs so we need to traverse the node hierarchy in Spoke to replace it in the child Armature node, otherwise Hubs will just read the original loop-animation component in the Armature node and use that.

src/editor/nodes/KitPieceNode.js
keianhzo: Updated.

Note that we still need to support the previous struct for backwards compatibility.

src/editor/nodes/ModelNode.js
keianhzo: After the naming update this has been replaced with a call to `getActiveClipIndices()` which traverses the animations and return the indices for the currently active clips. Basically what you suggested above.

keianhzo:
@robertlong I've also added the flag to exclude models from being combined.

robertlong:
@keianhzo I think these are the changes I'd like to make to this PR: https://github.com/mozilla/Spoke/compare/master...feature/multiple-animations

Let me know if these make sense.

src/editor/objects/Model.js
```diff
@@ -63,29 +64,58 @@ export default class Model extends Object3D {
...
     return clipOptions;
   }
 
-  get activeClip() {
-    return (this.model && this.model.animations && this.model.animations[this.activeClipIndex]) || null;
+  get activeClipNames() {
+    return this.activeClipIndices.filter(item => item.value >= 0).map(item => item.label);
+  }
+
+  getActiveClipIndices() {
```
johnshaughnessy, line 77: Should `getActiveClipIndices` return `this.activeClipIndices`? A bit confused on the naming / control flow here.

johnshaughnessy (commented):
I know this already went out :tada: -- but there seemed to have been a small bug so I left some comments here while trying to find it.

I think this PR has the fix but I'm not as familiar with this code as you so lmk if these comments/fix makes sense: https://github.com/mozilla/Spoke/pull/1036

src/editor/objects/Model.js
```diff
@@ -63,29 +64,58 @@ export default class Model extends Object3D {
...
-    clipOptions.unshift({ label: "None", value: -1 });
+    if (clipOptions.length == 0) {
+      clipOptions.unshift({ label: "None", value: -1 });
+    }
     return clipOptions;
   }
 
-  get activeClip() {
-    return (this.model && this.model.animations && this.model.animations[this.activeClipIndex]) || null;
+  get activeClipNames() {
```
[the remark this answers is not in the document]
keianhzo: Nope, I think that's a leftover from the PR WIP. I'll create a follow-up PR for cleanup.

src/editor/objects/Model.js
```diff
@@ -205,7 +250,7 @@ export default class Model extends Object3D {
     }
 
     this._src = source._src;
-    this.activeClipIndex = source.activeClipIndex;
+    this.activeClipIndices = source.activeClipIndices;
```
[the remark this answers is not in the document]
keianhzo: The deserialization is handled here:
https://github.com/keianhzo/Spoke/blob/eb46ee0564b945f1ba8f43a5ca4a43d3ae1c6e34/src/editor/nodes/ModelNode.js#L60
We don't handle the array conversion there but the select React component handles the conversion automatically, I'll update the deserialization with the guard to double check.

src/editor/nodes/ModelNode.js
```diff
@@ -48,17 +49,23 @@ export default class ModelNode extends EditorNodeMixin(Model) {
...
-          } else if (clip !== undefined && node.model && node.model.animations) {
-            // DEPRECATED: Old loop-animation component stored the clip name rather than the clip index
-            node.activeClipIndex = node.model.animations.findIndex(animation => animation.name === clip);
+          const { activeClipIndices } = loopAnimationComponent.props;
+
+          if (activeClipIndices !== undefined) {
+            node.activeClipIndices = activeClipIndices;
+          } else {
+            if (activeClipIndex !== undefined) {
+              node.activeClipIndices = activeClipIndex;
```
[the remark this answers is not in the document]
keianhzo: Yep, that's the one I mentioned in the previous comment, I'll update that one.

src/editor/nodes/KitPieceNode.js
```diff
@@ -28,17 +29,23 @@ export default class KitPieceNode extends EditorNodeMixin(Model) {
...
-          } else if (clip !== undefined && node.model && node.model.animations) {
-            // DEPRECATED: Old loop-animation component stored the clip name rather than the clip index
-            node.activeClipIndex = node.model.animations.findIndex(animation => animation.name === clip);
+          const { activeClipIndices } = loopAnimationComponent.props;
+
+          if (activeClipIndices !== undefined) {
+            node.activeClipIndices = activeClipIndices;
+          } else {
+            if (activeClipIndex !== undefined) {
+              node.activeClipIndices = activeClipIndex;
```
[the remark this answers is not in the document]
keianhzo: Yes, same case as for the ModelNode, React handles it but we sure make sure that it's always an array.

src/editor/nodes/ModelNode.js
```diff
@@ -48,17 +49,23 @@ export default class ModelNode extends EditorNodeMixin(Model) {
...
+            } else if (clip !== undefined && node.model && node.model.animations) {
+              // DEPRECATED: Old loop-animation component stored the clip name rather than the clip index
+              node.activeClipIndices = [node.model.animations.findIndex(animation => animation.name === clip)];
```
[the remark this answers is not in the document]
keianhzo: That would leave an empty array, I guess that's how it should be handled, right?

src/editor/objects/Model.js
keianhzo: Actually I've seen a bug reviewing this. We are not correctly restoring the animations, we are updating the `activeClipIndices` with the loop-animation stored indexes when deserializing. We need to convert that into a label/value struct so we can feed it to the React component. I'll address in the follow-up PR.

PR #1037, opened 2020-10-05
Fix migrations and add tests for animations

robertlong:


src/editor/nodes/KitPieceNode.js
```diff
@@ -34,18 +34,13 @@ export default class KitPieceNode extends EditorNodeMixin(Model) {
...
         if (loopAnimationComponent && loopAnimationComponent.props) {
-          const { clip, activeClipIndex } = loopAnimationComponent.props;
-          const { activeClipIndices } = loopAnimationComponent.props;
+          const { clip, activeClipIndices } = loopAnimationComponent.props;
 
-          if (activeClipIndices !== undefined) {
-            node.activeClipIndices = activeClipIndices;
+          if (clip !== undefined && node.model && node.model.animations) {
+            // DEPRECATED: Old loop-animation component stored the clip name rather than the clip index
+            node.activeClipIndices = [node.model.animations.findIndex(animation => animation.name === clip)];
```
johnshaughnessy, line 41: Is `-1` a valid clip index in case the model doesn't have a matching `animation.name`?

src/editor/objects/Model.js
```diff
@@ -246,7 +246,7 @@ export default class Model extends Object3D {
     }
 
     this._src = source._src;
-    this.activeClipIndices = source.activeClipIndices || (source.activeClipIndex && [source.activeClipIndex]) || [];
+    this.activeClipIndices = source.activeClipIndices;
```
johnshaughnessy, line 249: Ok, I'm glad this line went back. This was the weirdest feeling change I made. (And I don't have a good grasp on this code yet)

johnshaughnessy (approved):
Great, thanks for jumping on this

src/editor/nodes/KitPieceNode.js
robertlong: Yeah, that's definitely a bug, I'll fix it. Thanks!

PR #1038, opened 2020-10-07
Fix animation import issues

keianhzo:
This PR addresses some of the issues spotted by @johnshaughnessy here: https://github.com/mozilla/Spoke/pull/1029#pullrequestreview-502405581

- Update animation variables to be less confusing
- Correctly handle the animations deserialization. We we currently just populating `activeClipIndices` with the animation indices but that struct actually holds `SelectInput` items so we need to recreate those items based on the persisted animation indices.

### DIFF
--- a/src/editor/objects/Model.js
+++ b/src/editor/objects/Model.js
@@ -79,7 +79,7 @@ export default class Model extends Object3D {
           return { label: clip.name, value: item };
         });
     }
-    return null;
+    return [];
   }
 
   get activeClipIndices() {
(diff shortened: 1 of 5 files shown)


johnshaughnessy (approved):

src/editor/objects/Model.js
```diff
@@ -70,7 +70,19 @@ export default class Model extends Object3D {
...
+  getActiveItems(indices) {
+    if (this.model && this.model.animations) {
+      return indices
+        .filter(item => item >= 0 && this.model.animations[item])
+        .map(item => {
+          const clip = this.model.animations[item];
+          return { label: clip.name, value: item };
+        });
+    }
+    return null;
```
johnshaughnessy, line 82: Should this return an empty array?

keianhzo: React seems to handle that case but I have updated to avoid confusion.

johnshaughnessy (approved):

PR #1043, opened 2020-10-14
Add support for manual attribution update

### LINKED ISSUE #414: Need a field to allow manual attribution for acquired content
The use case:  I grabbed a model from Sketchfab, but edited it myself (Blender, etc.). Now when I add that model to Spoke, it doesn't have automatic attribution so I need a place to enter that info before publishing.

keianhzo:
closes #414 This PR adds support to show/edit attribution information in scene assets, specifically Model, Video, Image and Spawners. What this does right now:
- Update nodes attributions when media is resolved so in all cases where new media is added or updated (asset dropoff, node url update or from media gallery)
- All assets attribution is initially set based in the meta information returned by `resolveMedia` and then updated with the asset file info in case it's available (i.e. GLTF embedded asset info)
- Displays the attribution information in the Publishing dialog, one line per attribution element

It also fixes an issue with adding new source to audio/video nodes that threw an exception in Chrome an the updated element didn't work:
```
Cause:
    Failed to execute 'createMediaElementSource' on 'AudioContext': HTMLMediaElement already connected previously to a different MediaElementSourceNode.
```

We still need to solve the case where we can add per node GLTF attribution information based on https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_xmp/README.md Right now none of the sources support it but would be useful for exporting GLTF Spoke scenes while keeping all the attribution data. See https://github.com/mozilla/Spoke/issues/1044

### DIFF
--- a/src/api/PublishDialog.js
+++ b/src/api/PublishDialog.js
@@ -107,7 +107,18 @@ export default class PublishDialog extends Component {
         {contentAttributions && (
           <FormField>
             <label>Model Attribution:</label>
-            <p>{contentAttributions.map(a => `${a.name} by ${a.author}\n`)}</p>
+            <ul>
+              {contentAttributions.map(
+                (a, i) =>
+                  a.author &&
+                  a.title && (
+                    <li key={i}>
+                      <b>{`${a.title}`}</b>
+                      {(a.author && ` (by ${a.author})`) || ` (by Unknown)`}
+                    </li>
+                  )
+              )}
+            </ul>
           </FormField>
         )}
       </PreviewDialog>
--- a/src/editor/nodes/ModelNode.js
+++ b/src/editor/nodes/ModelNode.js
@@ -42,7 +42,8 @@ export default class ModelNode extends EditorNodeMixin(Model) {
         // Legacy, might be a raw string left over before switch to JSON.
         if (attribution && typeof attribution === "string") {
           const [name, author] = attribution.split(" by ");
-          node.attribution = { name, author };
+          node.attribution = node.attribution || {};
+          Object.assign(node.attribution, author ? { author: author } : null, name ? { title: name } : null);
         } else {
           node.attribution = attribution;
         }
@@ -115,15 +116,7 @@ export default class ModelNode extends EditorNodeMixin(Model) {
 
     const clonedScene = cloneObject3D(scene);
 
-    const sketchfabExtras = json.asset && json.asset.extras;
-
-    if (sketchfabExtras) {
-      const name = sketchfabExtras.title;
-      const author = sketchfabExtras.author ? sketchfabExtras.author.replace(/ \(http.+\)/, "") : "";
-      const url = sketchfabExtras.source || this._canonicalUrl;
-      clonedScene.name = name;
-      this.attribution = { name, author, url };
-    }
+    this.updateAttribution();
 
     return clonedScene;
   }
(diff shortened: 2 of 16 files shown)


src/api/PublishDialog.js
```diff
@@ -107,7 +107,18 @@ export default class PublishDialog extends Component {
         {contentAttributions && (
           <FormField>
             <label>Model Attribution:</label>
-            <p>{contentAttributions.map(a => `${a.name} by ${a.author}\n`)}</p>
+            <ul>
+              {contentAttributions.map(
+                (a, i) =>
+                  a.author &&
```
robertlong, line 113: We check for author down below as well. I think maybe we just want to check for title here?

src/editor/nodes/ModelNode.js
```diff
@@ -115,15 +116,7 @@ export default class ModelNode extends EditorNodeMixin(Model) {
 
     const clonedScene = cloneObject3D(scene);
 
-    const sketchfabExtras = json.asset && json.asset.extras;
-
-    if (sketchfabExtras) {
-      const name = sketchfabExtras.title;
-      const author = sketchfabExtras.author ? sketchfabExtras.author.replace(/ \(http.+\)/, "") : "";
```
robertlong, line 122: We may want to keep this string replacement to remove the author url from the author field. Otherwise you get an attribution that looks like this: 
Screen Shot 2021-01-12 at 4 35 46 PM

src/editor/objects/AudioSource.js
```diff
@@ -231,8 +231,12 @@ export default class AudioSource extends Object3D {
 
   async load(src) {
     await this.loadMedia(src);
-    this.audioSource = this.audioListener.context.createMediaElementSource(this.el);
```
robertlong, line 234: Can we add attributions to AudioNodes as well?

src/editor/nodes/SceneNode.js
```diff
@@ -637,7 +637,7 @@ export default class SceneNode extends EditorNodeMixin(Scene) {
       if (!attribution) return;
 
       if (attribution) {
-        const attributionKey = attribution.url || `${attribution.name}_${attribution.author}`;
```
robertlong, line 640: `getContentAttributions` is only gathering attributions for the `Model` nodes in the scene. We should check the `attribution` property instead of an explicit node type to see if it should be included.

robertlong (commented):
So sorry this took so long for me to review. There's a few changes to be made, but this is looking excellent!

src/editor/objects/AudioSource.js
keianhzo: This is actually unrelated to the attribution feature, it's related to what I mention in the PR notes: 

> It also fixes an issue with adding new source to audio/video nodes that threw an exception in Chrome an the updated element didn't work:
> 
> Cause:
>     Failed to execute 'createMediaElementSource' on 'AudioContext': HTMLMediaElement already connected previously 

But yep, I should have split that in another PR...

src/editor/nodes/ModelNode.js
keianhzo: Fixed

src/editor/nodes/SceneNode.js
keianhzo: Oops, missed that. Fixed.

src/api/PublishDialog.js
keianhzo: As long as we want to so attribution id just one is present. Don't know if it makes sense to show only the title without knowing the author or just the author without knowing to which asset is referring to.

keianhzo:
@robertlong Thanks for the review. I've fixed most of it. Still have doubts about the author/title check.

robertlong:
Looks good, excited to get it out there!

Issue #1045, opened 2020-10-14
Browser tab title shows scene name when back in Projects


j-conrad:
**Description:**
After editing a scene, returning 'Back to Projects' still shows the previously edited scene name in the browser tab title. This occurs in Firefox and Edge, although I'm guessing it would occur in any browser.

**Steps:**

1. Go to hubs.mozilla.com/spoke
2. Create a new scene or edit an existing one (note the name)
3. Save, then go to hamburger menu-->Back to Projects
4. Result: The browser tab still displays the name of that scene
![image](https://user-images.githubusercontent.com/31968496/96046475-67ed0780-0e28-11eb-81e0-020a578aa6c5.png)

**Details:**
OS: Windows 10
Browser: Firefox, Edge, maybe others

ThamiMemel:
I would like to work on this issue.

ThamiMemel:
I think this is caused by this HelmetJS code since there is no way to undo it right now.
https://github.com/mozilla/Spoke/blob/d787463ce4c778b4081a4f5c909e4e80463add29/src/ui/EditorContainer.js#L939-L942
I'm not sure this is the best approach to it, but this can be fixed by restoring the title in componentDidMount.
I'll send a PR with the change, please let me know if there is a better approach to this.

Softvision-GeluHaiduc:
I have verified this issue using Firefox and Chrome on Windows 10.
Now, the browser tab title shows the correct title name when back in projects.

PR #1047, opened 2020-10-23
Fix/title not changing

Yrkan:
Fixes: #1045
Added `document.title = configs.longName();` in` src/ui/projects/ProjectsPage.js`, this resets the page title when the user clicks back to projects (or returns to previous page).
Not sure it's the best approach for this issue.


robertlong:
Yeah, I think this approach is fine. Thanks for doing this @Yrkan :+1:

PR #1058, opened 2020-12-05
Allow toggling floating on objects from spawners

netpro2k:
Add the "Spawned object is floaty" option to object spawners. The default remains `true` which means objects are "floaty". Toggling it off will have the spawned object behave as if the "drop" button has been pressed. This is useful for things that are intended to be thrown like ping pong balls.


robertlong (approved):
LGTM, gave feedback in Discord on some naming, but otherwise looks great :+1:

PR #1070, opened 2021-01-12
Jenkins changes for QA builds

brianpeiris:
Changes the jenkins script to allow for building QA packages and adds a prompt for the QA promotion slack command.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,19 +1,5 @@
 import groovy.json.JsonOutput
 
-// From https://issues.jenkins-ci.org/browse/JENKINS-44231
-
-// Given arbitrary string returns a strongly escaped shell string literal.
-// I.e. it will be in single quotes which turns off interpolation of $(...), etc.
-// E.g.: 1'2\3\'4 5"6 (groovy string) -> '1'\''2\3\'\''4 5"6' (groovy string which can be safely pasted into shell command).
-def shellString(s) {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -43,14 +29,37 @@ pipeline {
           def slackURL = env.SLACK_URL
           def sentryDsn = env.SENTRY_DSN
           def gaTrackingId = env.GA_TRACKING_ID
+          def buildNumber = env.BUILD_NUMBER
+          def gitCommit = env.GIT_COMMIT
+          def jobName = env.JOB_NAME
+          def disableDeploy = env.DISABLE_DEPLOY
```
johnshaughnessy, line 35: Does this need a default value? Wondering if `${14}, ${15}` etc would fail in `hab-build-and-push` if it's an empty string. At some point we might want to consider moving away from positional arguments in our bash scripts (and instead used named arguments).

johnshaughnessy (approved):

Jenkinsfile
brianpeiris: Good catch. It will be a problem when we introduce new variables. I think it would be nice if we could use environment variables instead of expanding the script to support named arguments. We'll have to figure out if that's easy, but it should be doable.

brianpeiris: Actually, this is fine. It works out correctly because we've quoted the arguments below when calling the script.

PR #1077, opened 2021-01-14
Custom Navigation Meshes

### LINKED ISSUE #1075: Support for specifying a custom navmesh
It would be nice if you could directly specify the actual output navmesh and skip recast all together. This would probably be a setting on the "Floor Plan" element, where you could specify a glb, or maybe part of a glb to use.

robertlong:
You can now specify a model for a custom navigation mesh. In the Floor Plan element's properties panel change the floor plan mode to "Custom" and drag and drop a model into the "Custom Navmesh Url" field.

image

Fixes #1075


src/editor/nodes/FloorPlanNode.js
```diff
@@ -52,6 +60,20 @@ export default class FloorPlanNode extends EditorNodeMixin(FloorPlan) {
...
+
+    if (navMeshSrc) {
+      if (navMeshMode === NavMeshMode.Custom) {
+        loadAsync(
+          (async () => {
+            await node.load(navMeshSrc, onError);
+          })()
+        );
+      } else {
+        node._navMeshSrc = navMeshSrc;
```
netpro2k, line 73: You might consider just assigning this above right after navMeshMode like:
`node._navMeshSrc = navMeshSrc || ""`

to get rid of the need for the nested if, and make the default value more explicit.

src/editor/nodes/FloorPlanNode.js
```diff
@@ -52,6 +60,20 @@ export default class FloorPlanNode extends EditorNodeMixin(FloorPlan) {
     node.maxTriangles = maxTriangles || 1000;
     node.forceTrimesh = forceTrimesh || false;
 
+    node._navMeshMode = navMeshMode || NavMeshMode.Automatic;
+
+    if (navMeshSrc) {
+      if (navMeshMode === NavMeshMode.Custom) {
+        loadAsync(
+          (async () => {
+            await node.load(navMeshSrc, onError);
```
netpro2k, line 69: Do we need the wrapping function? I think `loadAsync(node.load(navMeshSrc, onError))`should do the same thing assuming `loadAsync` just expects a promise.

src/editor/nodes/FloorPlanNode.js
```diff
@@ -68,6 +90,88 @@ export default class FloorPlanNode extends EditorNodeMixin(FloorPlan) {
...
+      }
+
+      const geometry = mesh.geometry.clone(); // Clone in case the user reuses a mesh for the navmesh.
+      mesh.updateMatrixWorld();
+      geometry.applyMatrix(mesh.matrixWorld);
+
+      this.setNavMesh(new Mesh(geometry, new MeshBasicMaterial({ color: 0x0000ff, transparent: true, opacity: 0.2 })));
+
+      if (this.navMesh) {
+        this.navMesh.visible = this.editor.selected.indexOf(this) !== -1;
```
netpro2k, line 159: Is this needed? Guessing since you are emitting `selectionChanged` at the end of this function the state should already get updated.

src/ui/properties/FloorPlanNodeEditor.js
```diff
@@ -72,82 +88,93 @@ class FloorPlanNodeEditor extends Component {
...
+              smallStep={0.1}
+              mediumStep={1}
+              largeStep={10}
+              onChange={this.onChangeRegionMinSize}
+              unit="m²"
+            />
+          </>
+        ) : (
+          <InputGroup name="Custom Navmesh Url">
+            <ModelInput value={node.navMeshSrc} onChange={this.onChangeNavMeshSrc} />
```
netpro2k, line 175: Not new to this PR, just anything taking a model in general so not something we need to address in this PR, but we might want to think about how we could visually indicate somehow that this field can have models dragged into it from assets. Ideally there might even be a little button on the field that pops up an asset picker (similar to how it works in Unity)

netpro2k (commented):
Made some minor suggestions and discussed some possible future enhancements, but lgtm

src/editor/nodes/FloorPlanNode.js
robertlong: Yeah, it looks like `onSelect()` isn't called in response to the `selectionChanged` event :man_shrugging:

PR #1079, opened 2021-01-20
Enabled/Disabled Elements and Default Collapsed State

robertlong:
Screen Shot 2021-01-20 at 3 43 50 PM

Spoke elements can now be disabled, preventing them from being exported or included in nav mesh calculations. This can be helpful when testing scene elements or including additional content in your project for people remixing your scene.

Additionally, elements are now defaulted to the collapsed state. This should help declutter your scene and make groups more useful.

These two features can be used together to make make more advanced remixiable projects.


keianhzo (dismissed):
LGTM. The only thing I've seen is that even if the node is disabled we are still adding attribution information to the published scene. I guess we should only add that for the the enabled nodes.

Issue #1082, opened 2021-01-22
Featured scenes do not appear on the “New Project” page


Softvision-GeluHaiduc:
**[Affected Versions]:**
- Firefox Release 84.0.2

**[Affected Platforms]:**
- Windows 10x64

**[Prerequisites]:**
- A Hubs Cloud Instance is set up.
- Make sure to have some scenes imported, created and featured.
- Spoke is opened from a Hubs Cloud instance.

**[Steps to reproduce]:**
1. Click the "New Project" button.
2. Click the "Featured" button.
3. Observe the list.

**[Expected results]:**
- The list of featured projects is displayed.

**[Actual results]:**
- The list of featured projects is NOT displayed.

**[Notes]:**
- Attached a screen recording of the issue below
![featured projects](https://user-images.githubusercontent.com/51400741/105490906-2c32cb00-5cbe-11eb-88c4-e18b359ee657.gif)

misslivirose:
This is expected behavior. Featured scenes are different from Spoke project templates. Project templates are added separately from scene files in an instance.

Softvision-GeluHaiduc:
@misslivirose Could you please give me some more details on how to add project templates to the Hubs Cloud instance?

misslivirose:
As far as I understand it, changing the default projects in the scene editor requires deploying a custom deployment of Spoke to Hubs Cloud (the projects are in the /src/assets/templates directory). While I think some people have done it, I don't think we have an endorsed path for doing custom Spoke deployments that is well-tooled out yet, so I don't think it's in-scope right now for Hubs Cloud testing. (@johnshaughnessy, thoughts?)

johnshaughnessy:
I think the projects used to be pulled from `/src/assets/templates` as you said, but it looks like we changed that to pull `"featured-remixable"` scenes from `reticulum`:
https://github.com/mozilla/Spoke/blob/de2209b8f05e9ec73a6f0d3b59865a0ef6c310a9/src/ui/projects/CreateProjectPage.js#L115-L120

I can't tell if the templates in the repo are still used or if [maybe they should be removed](https://github.com/mozilla/Spoke/pull/1084).

There does not seem to be a convenient way to find these `featured-remixable` scene [listings?] in the admin panel, but I think to add a new one you would find a scene, feature it, and set to be remixable. I don't know the history here. Maybe @robertlong can add some details?

misslivirose:
[...]

PR #1084, opened 2021-01-25
Start removing unused templates

johnshaughnessy:
Not sure if this is correct... Are the templates in the Spoke repo used anymore?

See https://github.com/mozilla/Spoke/issues/1082#issuecomment-766979190 for context


robertlong:
Only the crater and tutorial templates were being used. I removed the others. This is ready to merge now.

PR #1086, opened 2021-01-27
Add support for creating scenes directly from GLBs

netpro2k:
Added support for publishing scenes directly from a GLB file. They are surfaced along side other projects on the projects page, but don't actually have a project associated with them. These GLB files are expected to be complete scenes including navmeshes, spawn points, collisions, etc, likely created using the Hubs Blender Exporter.

image

Its intentionally not surfaced as a primary call to action, and the intended users for this will likely already know they want it, and what its for.  Its likely we will eventually want to move this elsewhere and/or integrate it into our other tooling more directly, but we decided to surface it in Spoke as a starting point.

Todo:
- [x] Instead of using dummy projects query for scenes directly and merge the results on the projects page
- [x] test for regressions in existing publishing flows
- [x] merge https://github.com/mozilla/reticulum/pull/462

### DIFF
--- a/src/ui/projects/CreateProjectPage.js
+++ b/src/ui/projects/CreateProjectPage.js
@@ -123,7 +123,7 @@ export default function CreateProjectPage({ history, location }) {
                 </ProjectGridHeaderRow>
                 <ProjectGridHeaderRow>
                   <Button as={Link} to="/scenes/new">
-                    Upload GLB
+                    Import From Blender
                   </Button>
                   <Button as={Link} to="/projects/new">
                     New Empty Project
--- a/src/ui/projects/CreateProjectPage.js
+++ b/src/ui/projects/CreateProjectPage.js
@@ -122,6 +122,9 @@ export default function CreateProjectPage({ history, location }) {
                   <SearchInput placeholder="Search scenes..." value={params.q} onChange={onChangeQuery} />
                 </ProjectGridHeaderRow>
                 <ProjectGridHeaderRow>
+                  <Button as={Link} to="/scenes/new">
+                    Upload GLB
+                  </Button>
                   <Button as={Link} to="/projects/new">
                     New Empty Project
                   </Button>
--- a/src/ui/projects/CreateScenePage.js
+++ b/src/ui/projects/CreateScenePage.js
@@ -4,6 +4,7 @@ import { withApi } from "../contexts/ApiContext";
 import NavBar from "../navigation/NavBar";
 import Footer from "../navigation/Footer";
 import styled from "styled-components";
+import { Redirect } from "react-router-dom";
 
 import React, { useState, useCallback, useEffect } from "react";
 import { useHistory } from "react-router-dom";
@@ -21,6 +22,14 @@ const SceneUploadFormContainer = styled.div`
   flex-direction: row;
   background-color: ${props => props.theme.panel2};
   border-radius: 3px;
+`;
+
+const InfoBox = styled.div`
+  background-color: ${props => props.theme.panel2};
+  margin-top: 10px;
+  padding: 10px;
+  border-radius: 3px;
+  text-align: center;
 `;
 
 const ErrorMessage = styled.div`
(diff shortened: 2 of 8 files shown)


src/ui/projects/CreateScenePage.js
```diff
@@ -0,0 +1,366 @@
...
+  padding: 30px 30px;
+
+  label[type="button"] {
+    display: flex;
+    margin-bottom: 0;
+    margin-right: 5px;
+  }
+`;
+
+function CreateScenePage({ match, api }) {
```
robertlong, line 103: I think we need to check for authentication before routing to this page. Otherwise you can fill out the form and upon publishing, you will get a "Not Authenticated" error and lose data. If we check for authentication and then redirect back, users will not lose this data.

src/ui/projects/CreateProjectPage.js
```diff
@@ -122,6 +122,9 @@ export default function CreateProjectPage({ history, location }) {
                   <SearchInput placeholder="Search scenes..." value={params.q} onChange={onChangeQuery} />
                 </ProjectGridHeaderRow>
                 <ProjectGridHeaderRow>
+                  <Button as={Link} to="/scenes/new">
+                    Upload GLB
```
robertlong, line 126: I think we need to change the language of this button. Maybe "Import Project"? I think we may want to leave "GLB" out of it.

src/ui/projects/CreateScenePage.js
```diff
@@ -0,0 +1,366 @@
...
+  const [glbFile, setGlbFile] = useState(null);
+  const [thumbnailFile, setThumbnailFile] = useState(null);
+
+  const [thumbnailUrl, setThumbnailUrl] = useState(null);
+  const [sceneUrl, setSceneUrl] = useState(null);
+
+  const history = useHistory();
+
+  const openScene = useCallback(
+    function() {
```
robertlong, line 126: Typically, I'd use arrow functions for anonymous functions, especially for hooks. It's a stylistic thing, since the scope of `this` doesn't matter.

src/ui/projects/CreateScenePage.js
```diff
@@ -0,0 +1,366 @@
...
+      window.open(sceneUrl);
+    },
+    [sceneUrl]
+  );
+
+  useEffect(
+    function() {
+      async function doInitialLoad() {
```
robertlong, line 134: This doesn't need to be an arrow function since it's a named function. It does need a try/catch block or some `catch` handler on the async function invocation though.

src/ui/projects/CreateScenePage.js
```diff
@@ -0,0 +1,366 @@
...
+        setIsLoading(false);
+      }
+      doInitialLoad();
+    },
+    [match, api, setSceneInfo, setThumbnailUrl, setSceneId, setSceneUrl, setIsLoading]
+  );
+
+  const onChangeName = useCallback(
+    function(name) {
+      setSceneInfo({ ...sceneInfo, name });
```
robertlong, line 157: Rather than requiring `sceneInfo` as a dependency of this useCallback function, you can instead use `setSceneInfo(sceneInfo => ({ ...sceneInfo, name }))`.

More info on that form here: https://reactjs.org/docs/hooks-reference.html#functional-updates

src/ui/projects/CreateScenePage.js
```diff
@@ -0,0 +1,366 @@
...
+  const onChangeThumbnailFile = useCallback(
+    function([file]) {
+      if (thumbnailUrl && thumbnailUrl.indexOf("data:") === 0) {
+        URL.revokeObjectURL(thumbnailUrl);
+      }
+
+      setThumbnailFile(file);
+
+      // For preview
+      const reader = new FileReader();
```
robertlong, line 199: I think there's a race condition here should the component unmount before the reader has finished. It's an unlikely condition, but you can probably avoid it by using the FileReader in a `useEffect` hook.

```js
useEffect(() => {
  const reader = new FileReader();
  reader.onload = e => {
    setThumbnailUrl(e.target.result);
  };
  reader.readAsDataURL(thumbnailFile);
  
  return () => {
    reader.abort();
  };
}, [thumbnailFile, setThumbnailUrl])
```

src/ui/projects/CreateScenePage.js
```diff
@@ -0,0 +1,366 @@
...
+  );
+
+  const isNew = !sceneId;
+
+  const content = isLoading ? (
+    <ProgressBar />
+  ) : (
+    <>
+      <SceneUploadHeader>
+        <h1>{isNew ? "Publish Scene From GLB" : "Update GLB Scene"}</h1>
```
robertlong, line 249: Again, we should revisit this copy. I think this is clearer than the `Upload GLB` button though. Maybe we need some descriptive copy to go along with this and tell the user what is happening?

src/ui/projects/CreateScenePage.js
```diff
@@ -0,0 +1,366 @@
...
+          <FormField>
+            <FormField inline>
+              <label htmlFor="allowRemixing">
+                Allow{" "}
+                <a
+                  href="https://github.com/mozilla/Spoke/blob/master/REMIXING.md"
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  Remixing
```
robertlong, line 300: We'll want to update this copy to reflect the changes @emclaren has proposed.

src/ui/projects/ProjectGridItem.js
```diff
@@ -139,3 +151,23 @@ export default class ProjectGridItem extends Component {
...
+  const creatorAttribution = scene.attributions && scene.attributions.creator;
+  return (
+    <StyledProjectGridItem to={scene.url}>
+      <ThumbnailContainer>{scene.screenshot_url && <Thumbnail src={scene.screenshot_url} />}</ThumbnailContainer>
+      <TitleContainer>
+        <Col>
+          <h3>{scene.name}</h3>
+          {creatorAttribution && <p>{creatorAttribution}</p>}
+        </Col>
+        <Pill>GLB</Pill>
```
robertlong, line 165: Another spot where we want to check if the `GLB` copy is clear enough.

robertlong (commented):
Awesome work! Looks great, I saw a few things that could be cleaned up but generally very good. Especially great work on moving to React Hooks :+1: Thanks for doing that!

Can't wait for this to ship!

src/ui/projects/CreateProjectPage.js
netpro2k: Hmm, I think "project" would be confusing, since its not actually a spoke project you are importing (or creating). Its obviously weird since its in the projects flow and surfaced on the projects page right now, but I do think mentioning GLB seems important. Maybe something like "create project from GLB"?

src/ui/projects/CreateScenePage.js
netpro2k: Makes sense. Do we have something to mark routes as authenticated or should it just use the history api to redirect?

src/ui/projects/CreateScenePage.js
netpro2k: I tend to only do it when the binding actually matters, but will change it to match the project style.

src/ui/projects/CreateScenePage.js
netpro2k: As per discussion keeping it this way on this button.

src/ui/projects/CreateScenePage.js
netpro2k: Took a look around and wasn't able to find this copy. Since it hasn't been merged anyway it makes sense for this to be consistent with the publish dialog. We will just need to make sure to update this in the PR where we change the other publish flow copy.

PR #1089, opened 2021-01-28
Use template literals- Best Practices.

hakerbaya:


robertlong:
Thanks @hakerbaya :+1:

Issue #986, opened 2020-06-30
Spot Light intensity uses degrees rather than candela


robertlong:
We need to change the unit of the input.

sardug10:
Hey, I am new to open source contributions. Is this a good issue to start with?

robertlong:
Yeah, definitely a good one. It just needs to be changed here:
https://github.com/mozilla/Spoke/blob/master/src/ui/properties/SpotLightNodeEditor.js#L61

sardug10:
> Yeah, definitely a good one. It just needs to be changed here:
> https://github.com/mozilla/Spoke/blob/master/src/ui/properties/SpotLightNodeEditor.js#L61

Hey, instead of unit="°" , I just have to change it to unit="cd". Is it?😃

PR #1094, opened 2021-02-02
Updated unit of Spot Light Intensity

sardug10:
Changed Spot Light Intensity Unit from 'Degrees' to 'Candela'  [here.](https://github.com/mozilla/Spoke/blob/master/src/ui/properties/SpotLightNodeEditor.js#L61)

It fixes [#986](https://github.com/mozilla/Spoke/issues/986)


robertlong:
Awesome, thanks!

PR #1101, opened 2021-02-12
Make Images and Videos Linkable

robertlong:
You can now add links to Videos and Images. Just fill out the href field with a url and your media when someone clicks on the media they will open up a new tab with that page.

image

Depends on:
https://github.com/mozilla/hubs/pull/3904


src/ui/properties/ImageNodeEditor.js
```diff
@@ -38,6 +40,9 @@ export default function ImageNodeEditor(props) {
       <InputGroup name="Billboard" info="Image always faces user in Hubs.">
         <BooleanInput value={node.billboard} onChange={onChangeBillboard} />
       </InputGroup>
+      <InputGroup name="Link Href" info="Allows the image to function as a link for the given url.">
```
netpro2k, line 43: Should we hide this if projection is not set to flat since we don't export in that case?

src/ui/properties/VideoNodeEditor.js
```diff
@@ -27,6 +29,9 @@ export default function VideoNodeEditor(props) {
       <InputGroup name="Billboard" info="Video always faces user in Hubs. Does not billboard in Spoke.">
         <BooleanInput value={node.billboard} onChange={onChangeBillboard} />
       </InputGroup>
+      <InputGroup name="Link Href" info="Allows the video to function as a link for the given url.">
+        <StringInput value={node.href} onChange={onChangeHref} />
```
netpro2k, line 33: Same here

netpro2k (commented):
lgtm. We might want to hide the link field in cases where it will not be exported to avoid confusion.

PR #1130, opened 2021-04-15
Allow passing through components with node refferences

netpro2k:
This adds support for importing GLTFs that contain Hubs components with node references. The GLTF index in values of specific components (defined in `HUBS_NODEREF_COMPONENTS` for now) is remapped to a UUID of the target node at import time. At export time this value is remapped back to the index of the same node in the final exported GLTF. This piggybacks on the existing method used in Spoke for exporting these references.  We should be able to apply a similar technique for other sorts of internal references (materials, images, etc) as needed. We plan to replace the GLTFLoader/GLTFExporter soon but this should be easily ported to the new plugin API.

Separately, also allows the GLTF exporter to export cameras so they are correctly passed through.

These are both used in the new `video-texture-target`/`video-texture-soruce` components https://github.com/mozilla/hubs/pull/4126

### DIFF
--- a/src/editor/gltf/GLTFExporter.js
+++ b/src/editor/gltf/GLTFExporter.js
@@ -1262,6 +1262,47 @@ class GLTFExporter {
   }
 
   /**
+   * Process camera
+   * @param  {THREE.Camera} camera Camera to process
+   * @return {Integer}      Index of the processed mesh in the "camera" array
+   */
+  processCamera(camera) {
+    if (!this.outputJSON.cameras) {
+      this.outputJSON.cameras = [];
+    }
+
+    const isOrtho = camera.isOrthographicCamera;
+
+    const gltfCamera = {
+      type: isOrtho ? "orthographic" : "perspective"
+    };
+
+    if (isOrtho) {
+      gltfCamera.orthographic = {
+        xmag: camera.right * 2,
+        ymag: camera.top * 2,
+        zfar: camera.far <= 0 ? 0.001 : camera.far,
+        znear: camera.near < 0 ? 0 : camera.near
+      };
+    } else {
+      gltfCamera.perspective = {
+        aspectRatio: camera.aspect,
+        yfov: _Math.degToRad(camera.fov),
+        zfar: camera.far <= 0 ? 0.001 : camera.far,
+        znear: camera.near < 0 ? 0 : camera.near
+      };
+    }
+
+    if (camera.name !== "") {
+      gltfCamera.name = camera.type;
+    }
+
+    this.outputJSON.cameras.push(gltfCamera);
+
+    return this.outputJSON.cameras.length - 1;
+  }
+
+  /**
    * Process Object3D node
    * @param  {THREE.Object3D} node Object3D to processNode
    * @return {Integer}      Index of the node in the nodes list
@@ -1312,6 +1353,8 @@ class GLTFExporter {
       if (mesh !== null) {
         gltfNode.mesh = mesh;
       }
+    } else if (object.isCamera) {
+      gltfNode.camera = this.processCamera(object);
     }
 
     if (object.isSkinnedMesh) {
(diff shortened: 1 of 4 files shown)


src/editor/gltf/GLTFLoader.js
```diff
@@ -1990,6 +1995,22 @@ class GLTFLoader {
...
+          for (const [componentName, componentProps] of Object.entries(components)) {
+            for (const [propName, propValue] of Object.entries(componentProps)) {
+              if (
+                HUBS_NODEREF_COMPONENTS[componentName] &&
+                HUBS_NODEREF_COMPONENTS[componentName].indexOf(propName) !== -1
+              ) {
+                this.getDependency("node", propValue).then(node => {
+                  node.userData.MOZ_spoke_uuid = node.uuid;
+                  componentProps[propName] = { __gltfIndexForUUID: node.uuid };
+                });
```
takahirox, line 2009: `getDependency()` is async method so no guarantee that it is resolved by the end of loader's parse (= by loader's `onLoad()` callback is called). I'm not familiar with Spoke yet but I think it would be problematic, wouldn't it?

src/editor/gltf/GLTFExporter.js
```diff
@@ -1261,6 +1261,47 @@ class GLTFExporter {
...
+      gltfCamera.perspective = {
+        aspectRatio: camera.aspect,
+        yfov: _Math.degToRad(camera.fov),
+        zfar: camera.far <= 0 ? 0.001 : camera.far,
+        znear: camera.near < 0 ? 0 : camera.near
+      };
+    }
+
+    if (camera.name !== "") {
+      gltfCamera.name = camera.type;
```
takahirox, line 1297: Is this intentional that you save type as name if name is not empty?

takahirox (commented):
I'm not familiar with Spoke and the data layout yet so it's hard for me to determine whether this change is good or not so I have only commented some Three.js stuffs so far.

src/editor/gltf/GLTFLoader.js
netpro2k: Yeah its definitely not ideal. Doing this the "right" way would mean refactoring GLTF loader some more and since we are going to replace it soon it doesn't seem worth it. This info is only used on export so the promise is going to be well resolved by then.

src/editor/gltf/GLTFExporter.js
netpro2k: This code was copied from the upstream GLTFExporter at the time. It was removed earlier in a Spoke commit so I just brought it back.

takahirox: I think this should be fixed in the official `GLTFExporter` side.

src/editor/gltf/GLTFLoader.js
robertlong: Given how we're loading glTFs, loading and awaiting the whole scene, I think it'll be fine. If you're able to get this to turn into a race condition, you're going to have other problems. Like the model just isn't in the output.

src/editor/gltf/GLTFLoader.js
```diff
@@ -192,6 +192,11 @@ const GLB_HEADER_MAGIC = "glTF";
 const GLB_HEADER_LENGTH = 12;
 const GLB_CHUNK_TYPES = { JSON: 0x4e4f534a, BIN: 0x004e4942 };
 
+const HUBS_NODEREF_COMPONENTS = {
```
robertlong, line 195: I'd probably move this map somewhere else and import it into this GLTFLoader file. I think it'll be easier to maintain that way. Ideally it'll eventually be a glTF plugin, so maybe put it where you'll eventually write that in a separate file in this glTF folder.

robertlong (approved):
Looks good! Excited to get this working!

src/editor/gltf/GLTFLoader.js
takahirox: I would like to suggest to add a comment here about why we don't need to wait. Otherwise, other devs might be confused whey they read this part.

netpro2k: Missed this before pushing the PR, but added a comment now.

PR #1156, opened 2021-07-09
Audio Zones

keianhzo:
- Audio zones support
- Refactor to use Audio params

Related https://github.com/mozilla/hubs/pull/4399
Related https://github.com/MozillaReality/hubs-blender-exporter/pull/36


src/ui/properties/VideoNodeEditor.js
```diff
@@ -8,7 +8,7 @@ import StringInput from "../inputs/StringInput";
 import { VideoProjection } from "../../editor/objects/Video";
 import VideoInput from "../inputs/VideoInput";
 import { Video } from "styled-icons/fa-solid/Video";
-import AudioSourceProperties from "./AudioSourceProperties";
+import AudioParamsProperties from "./AudioParamsProperties";
```
netpro2k, line 11: We discussed this a bit yesterday.. I wonder if `audio-source' is a better name for this than `audio-params`

src/editor/nodes/VideoNode.js
```diff
@@ -18,22 +18,44 @@ export default class VideoNode extends EditorNodeMixin(Video) {
...
       rolloffFactor,
       refDistance,
       maxDistance,
       coneInnerAngle,
       coneOuterAngle,
-      coneOuterGain,
-      projection
-    } = json.components.find(c => c.name === "video").props;
+      coneOuterGain;
+    const audioParamsComp = json.components.find(c => c.name === "audio-params");
```
netpro2k, line 33: I wonder if this warrants a version bump of the spoke scene version and  "migration". See:  https://github.com/mozilla/Spoke/blob/dev/src/editor/nodes/SceneNode.js#L196 ... Not sure what our criteria are for this.

src/editor/nodes/AudioZoneNode.js
```diff
@@ -0,0 +1,125 @@
...
+        refDistance: this.refDistance,
+        maxDistance: this.maxDistance,
+        coneInnerAngle: this.coneInnerAngle,
+        coneOuterAngle: this.coneOuterAngle,
+        coneOuterGain: this.coneOuterGain
+      }
+    });
+  }
+
+  prepareForExport() {
```
netpro2k, line 96: So.. This looks like what we discussed we wanted yesterday, where no mesh would actually get exported, but I am a bit surprised how this was working currently since the hubs-side code wanted the object's bounding box...

netpro2k (commented):
This lgtm though I am surprised because I don't see new commits and this already seems to be doing the behavior we discussed where Audio Zones will be empty nodes... How was this working before? Also I made note about migrations in case you hadn't seen that pattern in Spoke, though I don't really have a strong opinion on if this warrants a version bump or not.

src/editor/nodes/AudioZoneNode.js
keianhzo: The are applying the world transform to the BBAA in the client here:
https://github.com/mozilla/hubs/blob/ad8e269ceaac57e28b10cb781c755a74ac0a1df8/src/components/audio-zone.js#L68-L73 
```
  getBoundingBox: (function() {
    const bbaa = new THREE.Box3();
    return function() {
      return bbaa.copy(this.debugMesh.geometry.boundingBox).applyMatrix4(this.el.object3D.matrixWorld);
    };
  })(),
```
We are also adding the debug mesh to the root object node so the transform is applied to the child mesh and the transform is the exported GLTF so the debug object has the right transform.

src/editor/nodes/VideoNode.js
keianhzo: Yep, I think think you are right, I''m going to add a migration step. It's better than adding backwards compatibility and a good practice.

src/editor/nodes/AudioZoneNode.js
fradomnt: Thanks!!! amazing work!

Issue #1160, opened 2021-07-22
Cannon configure sounds from dynamic videos


maucascemiro:
I've been workin on a virtual exposition for 40 days, and out of the blue i  do not have acces to some features like the AUTO PLAY box when I place a dynamic youtube video. Another problem is the fact that I cant configure my videos in order to only be able to hear it near. Pls help

keianhzo:
This PR https://github.com/mozilla/Spoke/pull/1161 will restore this properties. Make sure the you edit the scene and set those values back to their original state if you published that scene.

keianhzo:
This has been merged so if you open your scene in Spoke now, you should see the auto-play checkbox again and you can check it for the media elements that were unchecked.

PR #1161, opened 2021-07-29
Restore lost media params

keianhzo:
During the Audio zones PR some media properties were accidentally removed from the audio and video nodes.

This PR restores those params

Fixes https://github.com/mozilla/Spoke/issues/1160


maucascemiro:
hey thank you for the feed back. Im still having some problems with audio in my scenes.. most of the times the scene opens with all audio from all videos open as if i never configured it, but all my videos are inside audio zones and properly configured. this bug does not hapens everytime I open the scene, but most of the times. (im sorry for bad english)

maucascemiro:
Im not sure how to aply your fix to my scene.

brianpeiris (approved):
LGTM.

PR #1195, opened 2021-11-23
Troika Text node

jameskane05:
This is the Spoke PR for creating, editing, copying, deleting and exporting Troika Text nodes, based on the troika-three-text package (https://www.npmjs.com/package/troika-three-text). This package has a wide variety of styling options, plus the advantage of being able to display 3D text consistently in Spoke (pure three.js) and Hubs (A-Frame), making it a superior choice for content management purposes versus A-Frame’s `<a-text>` nodes or Mozilla’s hubs-text.js method. There should also be interesting programmatic update use cases for this node.

Overall, 3D artists should not be responsible for _all_ content management, so moving text management out of Blender and into Spoke should help - just like in traditional web dev, you'd build a CMS so that your graphic designer doesn't have to change out copy every single time.

To test:
1. Switch to `feature/troika-text` branch and `yarn install`
2. Create a new scene and add a Troika Text node, customizing the data as desired
3. Save the scene, reload, and see the Troika Text node has persisted
4. Copy the object to see values persist from source object to the copy
5. Delete an object to see text nodes removed from the scene and disposed of
6. Publish the scene and load with this Hubs PR: https://github.com/mozilla/hubs/pull/4855

image

KNOWN ISSUE: `troika-three-text` has a "Font URL" field, which accepts links to .ttf, .otf and .woff files, downloads them at runtime and converts them to SDF glyphs. Users might expect to be able to upload those file types directly to Spoke, however currently there are both Spoke-side and server-side file type upload limitations that prevents this. For now you can only use remotely-hosted fonts (and watch out for CORS errors), and they cannot be uploaded to Spoke.

### DIFF
--- a/src/editor/nodes/TroikaTextNode.js
+++ b/src/editor/nodes/TroikaTextNode.js
@@ -1,0 +1,217 @@
+import { Color, Object3D } from "three";
+import EditorNodeMixin from "../../../../src/editor/nodes/EditorNodeMixin";
+import { Text } from "troika-three-text";
+
+export default class TroikaTextNode extends EditorNodeMixin(Object3D) {
+  static legacyComponentName = "troika-text";
+
+  static nodeName = "Troika Text";
+
+  constructor(editor) {
+    super(editor);
+    this.editor = editor;
+    this.scene = editor.scene;
+    this.text = "Text";
+    this.anchorX = "left";
+    this.anchorY = "top";
+    this.color = new Color();
+    this.curveRadius = 0;
+    this.depthOffset = 0;
+    this.direction = "auto";
+    this.fillOpacity = 1;
+    this.font;
+    this.fontSize = 10;
+    this.textAlign = "left";
+    this.letterSpacing = 0;
+    // this.clipRect = "";  // Todo, implement later
+    this.lineHeight = 1;
+    this.outlineBlur = 0;
+    this.outlineColor = new Color();
+    this.outlineOffsetX = 0;
+    this.outlineOffsetY = 0;
+    this.outlineOpacity = 1;
+    this.outlineWidth = 0;
+    this.overflowWrap = "normal";
+    this.strokeColor = new Color();
+    this.strokeOpacity = 1;
+    this.strokeWidth = 0;
+    this.textAlign = "left";
+    this.textIndent = 0;
+    this.whiteSpace = "normal";
+    this.maxWidth = 9999; // The serialize method can't seem to handle a value of Infinity, just feed it ~10k meters
+    this.troikaText = new Text();
+  }
+
+  onChange(change) {
+    // change is undefined when serialize returns - but whole node has been updated
+    // change will be named if coming from a UI input
+    if (change === "position") this.troikaText.position.copy(this.position);
+    else if (change === "scale") this.troikaText.scale.copy(this.position);
+    else if (change === "rotation") this.troikaText.rotation.copy(this.rotation);
+    else if (change === "color" || change === "strokeColor" || change === "outlineColor") {
+      this.troikaText[change] = new Color(this[change]);
+    } else this.troikaText[change] = this[change];
+    this.troikaText.sync();
+  }
+
+  onRemove() {
+    this.scene.remove(this.troikaText);
+    this.troikaText.dispose();
+  }
+
+  onAdd() {
+    this.troikaText = new Text();
+    this.troikaText.text = this.text;
+    this.troikaText.anchorX = this.anchorX;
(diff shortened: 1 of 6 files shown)


jameskane05:
Updated this PR with a couple bug fixes we discovered in production. Needed to add the render helper directly to the node vs. to the scene in order for normal node functionality. Also fixed value copy issues.

src/editor/nodes/TroikaTextNode.js
```diff
@@ -0,0 +1,219 @@
+// This is a port of https://github.com/protectwise/troika/tree/master/packages/troika-three-text
+// that has been made functional and restyled for Mozilla Hubs and Spoke (hubs.mozilla.com)
+// by @jamesckane at Paradowski Creative (paradowski.com)
+
+import { Color, Object3D } from "three";
+import EditorNodeMixin from "../../../../src/editor/nodes/EditorNodeMixin";
```
keianhzo, line 6: I think this should be jut ./EditorNodeMixin

src/editor/nodes/TroikaTextNode.js
```diff
@@ -0,0 +1,219 @@
...
+import { Text } from "troika-three-text";
+
+export default class TroikaTextNode extends EditorNodeMixin(Object3D) {
```
keianhzo, line 9: Maybe it we could just inherit from Text directly instead of Object3D and save the extra node?

src/editor/nodes/TroikaTextNode.js
```diff
@@ -0,0 +1,219 @@
...
+    this.text = "Text";
+    this.anchorX = "left";
+    this.anchorY = "top";
+    this.color = new Color();
+    this.curveRadius = 0;
+    this.depthOffset = 0;
+    this.direction = "auto";
+    this.fillOpacity = 1;
+    this.font;
+    this.fontSize = 10;
```
keianhzo, line 27: I feel like 10 is quite a big font size for the default value. What about 1?

src/editor/nodes/TroikaTextNode.js
```diff
@@ -0,0 +1,219 @@
...
+    // change is undefined when serialize returns - but whole node has been updated
+    // change will be named if coming from a UI input
+    if (change === "position" || change === "scale" || change === "rotation") return;
+    else if (change === "color" || change === "strokeColor" || change === "outlineColor") {
+      this.troikaText[change] = new Color(this[change]);
+    } else this.troikaText[change] = this[change];
+    this.troikaText.sync();
+  }
+
+  onRemove() {
```
keianhzo, line 59: We can probably avoid `onAdd` and `onRemove` if we inherit from `Text`.

src/ui/properties/TroikaTextNodeEditor.js
```diff
@@ -0,0 +1,361 @@
...
+// that has been made functional and restyled for Mozilla Hubs and Spoke (hubs.mozilla.com)
+// by @jamesckane at Paradowski Creative (paradowski.com)
+
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import NodeEditor from "../../../../src/ui/properties/NodeEditor";
+import InputGroup from "../../../../src/ui/inputs/InputGroup";
+import StringInput from "../../../../src/ui/inputs/StringInput";
+import { AlignCenter } from "styled-icons/fa-solid/AlignCenter";
+import SelectInput from "@src/ui/inputs/SelectInput";
```
keianhzo, line 11: Maybe there is some webpack module that you are using for `@` that's not been commited?

src/ui/properties/TroikaTextNodeEditor.js
```diff
@@ -0,0 +1,361 @@
...
+          value={node.letterSpacing}
+          onChange={value => {
+            this.onChangeProperty("letterSpacing", value);
+          }}
+          unit="em"
+        />
+
+        <NumericInputGroup
+          name="Line Height"
+          info="Sets the height of each line of text. Can either be 'normal' which chooses a reasonable height based on the chosen font's ascender/descender metrics, or a number that is interpreted as a multiple of the fontSize."
```
keianhzo, line 215: I believe `normal` is not supported here.

keianhzo (changes requested):
This is great! Just a few nits but this is mostly awesome and it's going to be really useful.
Thanks so much for this.

src/editor/nodes/TroikaTextNode.js
jameskane05: These are now handled by the `this.helper` convention used throughout the Spoke project.

PR #1196, opened 2021-11-29
Fix the What's New page

brianpeiris:
This PR updates the public token used for the What's New page. It also changes the sort order of the pull request query, so that old PRs with new comments don't get sorted to the top.


brianpeiris:
Need to amend this PR. Will request review again shortly.

brianpeiris:
As implied in the code, the token used here only provides read-only access to information that is already public. It has no OAuth scopes enabled, so it is safe to include in the repo.

keianhzo (approved):
LGTM

PR #1197, opened 2021-12-02
Add dockerfile, scripts, and github workflow for Turkey ops

netpro2k:
This is essentially exactly the same as Hubs, except Spoke uses yarn instead of npm.

I used `turkeyCfg_base_asset_path` instead of constructing a url from `DOMAIN` and `SUB_DOMAIN` as that seems more correct. If there wasn't a reason for this we should do this in Hubs as well.

I noticed in the Hubs run.sh we have a sort of fake http server for liveness/readyness checks. Any reason not to just request index.html from the actual nginx server? Simpler and actually represents the health of the pod.

I have only tested this insofar as the substitution in index.html looks sane and nginx serves it up on the expected port/path. Still need to try actually pointing a reticulum at it.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git
+node_modules
--- a/.github/workflows/spoke-RetPageOrigin.yml
+++ b/.github/workflows/spoke-RetPageOrigin.yml
@@ -1,0 +1,15 @@
+name: spoke
+on:
+  push:
+    branches:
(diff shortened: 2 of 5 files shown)


tanfarming:
looks great thanks for the help lol

tanfarming (approved):

brianpeiris:
Looks like this was merged in #1212

PR #1212, opened 2022-02-02
Feature/turkey dockerr

tanfarming:


scripts/docker/run.sh
```diff
@@ -0,0 +1,13 @@
...
+find /www/spoke/ -type f -name *.css -exec sed -i "s~{{rawspoke-base-assets-path}}~${turkeyCfg_base_assets_path}~g" {} \;
+
+anchor="<!-- DO NOT REMOVE\/EDIT THIS COMMENT - META_TAGS -->"
+for f in /www/spoke/pages/*.html; do
+    for var in $(printenv); do
+    var=$(echo $var | cut -d"=" -f1 ); prefix="turkeyCfg_";
+    [[ $var == $prefix* ]] && sed -i "s/$anchor/ <meta name=\"env:${var#$prefix}\" content=\"${!var//\//\\\/}\"\/> $anchor/" $f;
+    done
+done
+nginx -g "daemon off;"
```
brianpeiris, line 13: I think we might have mentioned this in the Hubs docker PR, but ideally we would just serve these static assets from S3 instead of actually running nginx continuously in a container.

PR #1233, opened 2022-03-12
Add Mirror component

juunini:
# Doing

image

- Add Mirro component
    - https://github.com/mozilla/hubs/pull/5225

# Issue

~~Copy and paste mirror, the pasted mirror's color is not changed.~~
~~But when publish scene, in hubs color is correct.~~
~~I don't know how to fix it. plz help.~~
I fix it. clear.

# Description

This feature is only apply plane mirror.
But, I'll add shape and texture properties in future.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -50,6 +50,8 @@ import ScenePreviewCameraNode from "./editor/nodes/ScenePreviewCameraNode";
 import ScenePreviewCameraNodeEditor from "./ui/properties/ScenePreviewCameraNodeEditor";
 import AudioZoneNode from "./editor/nodes/AudioZoneNode";
 import AudioZoneNodeEditor from "./ui/properties/AudioZoneNodeEditor";
+import MirrorNode from "./editor/nodes/MirrorNode";
+import MirrorNodeEditor from "./ui/properties/MirrorNodeEditor";
 
 import MediaFrameNode from "./editor/nodes/MediaFrameNode";
 import MediaFrameNodeEditor from "./ui/properties/MediaFrameNodeEditor";
@@ -93,6 +95,7 @@ export function createEditor(api, settings) {
   editor.registerNode(ScenePreviewCameraNode, ScenePreviewCameraNodeEditor);
   editor.registerNode(MediaFrameNode, MediaFrameNodeEditor);
   editor.registerNode(AudioZoneNode, AudioZoneNodeEditor);
+  editor.registerNode(MirrorNode, MirrorNodeEditor);
 
   editor.registerSource(new ElementsSource(editor));
   editor.registerSource(new MyAssetsSource(editor));
--- a/src/editor/nodes/MirrorNode.js
+++ b/src/editor/nodes/MirrorNode.js
@@ -1,0 +1,48 @@
+import Mirror from "../objects/Mirror";
+import EditorNodeMixin from "./EditorNodeMixin";
+
+export default class MirrorNode extends EditorNodeMixin(Mirror) {
+  static componentName = "mirror";
+
+  static nodeName = "Mirror";
+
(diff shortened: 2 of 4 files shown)


johnshaughnessy (approved):
Works great!

johnshaughnessy:
@juunini Do you think we should warn users about the potential performance impact of mirrors by adding a "Mirrors" section to the [Performance Check Dialog](https://github.com/mozilla/Spoke/blob/6ec35b73b644b2fb0e8221426f6256d9b7be43a6/src/ui/dialogs/PerformanceCheckDialog.js#L93)?

perf dialog

johnshaughnessy:
I've edited the PR summary so that this appears in Spoke's "what's new" page: https://hubs.mozilla.com/spoke/whats-new

juunini:
@johnshaughnessy
Umm... I don't have any idea to check performance impact of mirror.
Do you have any idea to check performance impact?

johnshaughnessy:
@juunini I did not do a formal performance check, but I am assuming there will be an impact because the `Reflector` does separate render of the scene. I'll ask @takahirox what he thinks.

@takahirox 
1. Do you think we should do anything to warn users about the performance impact of mirrors?
2. Do you think we should do anything in the client to reduce the performance impact of mirrors (especially for low-powered devices like mobile and tablets)? (See https://github.com/mozilla/hubs/pull/5225#pullrequestreview-908004513 ). 

As an example, when you spawn an in-game `camera-tool`, we only update the viewfinder `VIEWFINDER_FPS` (6) times per second unless holding, snapping, recording video with it:

```js
    // Always draw held, snapping, or recording camera viewfinders with a decent framerate
    if (
      (isHolding || this.data.isSnapping || this.videoRecorder) &&
      performance.now() - this.lastUpdate >= 1000 / (this.videoRecorder ? VIDEO_FPS : VIEWFINDER_FPS)
    ) {
      this.updateRenderTargetNextTick = true;
    }
```
https://github.com/mozilla/hubs/blob/1938d7a5daf3801a5f94ff67c0fa29e3f933b294/src/components/camera-tool.js#L532-L538

johnshaughnessy:
@keianhzo I did not notice that this PR failed a "clone" test. What does that mean? How do we fix it? https://app.circleci.com/pipelines/github/mozilla/Spoke/1372/workflows/c5d98eb2-8e42-4dfe-a2c2-ae209f06af5e/jobs/1562

juunini:
I don't have to check performance idea.
But I have idea what reduce performance for low power device.

- reduce view finder fps
- reduce texture width and texture height property of mirror (aliased)

keianhzo:
> @keianhzo I did not notice that this PR failed a "clone" test. What does that mean? How do we fix it? https://app.circleci.com/pipelines/github/mozilla/Spoke/1372/workflows/c5d98eb2-8e42-4dfe-a2c2-ae209f06af5e/jobs/1562

Yep, it's not recursively copying the children so the cloning test fails as that test adds a child to test the recursive cloning. I've also added the color initialization as otherwise the uniform color is not initialized and the initial color is not set and also makes the clone test to fail.

https://github.com/mozilla/Spoke/pull/1234

takahirox:
> I am assuming there will be an impact because the Reflector does a separate render of the scene every tic.

Probably yes.

> 1. Do you think we should do anything to warn users about the performance impact of mirrors?

Depending how big the performance impact is. Currently we don't have systematic performance tests. Maybe it won't be too late to start to warn if we notice the slowness or users report the bad performance.

> 2. Do you think we should do anything in the client to reduce the performance impact of mirrors (especially for low-powered devices like mobile and tablets)? 

What I came up with so far is reducing the texture (render target) resolution for mobile devices or lower material quality mode.

PR #1255, opened 2022-08-05
Update creating-custom-elements

jeus0630:
## DOING
- Update `THREE.Math.degToRad` to `THREE.MathUtils.degToRad`

## WHY
- It doesn't work with `Math`, it is `MathUtils`


juunini:
NOTE: Three.js update issue.
https://github.com/mrdoob/three.js/releases/tag/r113
Rename to MathUtils. https://github.com/mrdoob/three.js/pull/18270 (@Mugen87)

PR #1258, opened 2022-09-12
Prepare for local orchestration

bryanenders:
Why
---
An “Invalid Host header” error message appears in the browser when Hubs is orchestrated with Docker Compose

What
----
* Set sufficient self-signed certificate key size for dev
* Allow internal hostname to be supplied by OS var


brianpeiris:
You can ignore the CircleCI build issue here. It's unrelated to your changes, and something we need to fix separately.

PR #1293, opened 2024-03-19
Fix publishing scenes with no models i.e. the e is not iterable error.

### LINKED ISSUE #1137: Can't publish a scene with no models in it
Sometimes I want to publish scenes with either nothing in it or just a skybox, but the scene will not publish and shows an error of "e is not iterable".  The reason I want scenes with no objects is usually so that I can have a stable baseline for testing performance of objects added into a room, or I'm testing skyboxes, stuff like that.  This appears to be a fairly recent regression, markun mentioned it in the spoke discord channel on April 20th, but I think it was working not too long before that.

Exairnous:
Fixes #1137
Skip trying to remap the node refs for materials if there aren't any materials in the scene. Add the length of the JSON chunk to the header when there is no binary buffer so that a valid GLB file can be generated.

### DIFF
--- a/src/editor/Editor.js
+++ b/src/editor/Editor.js
@@ -424,7 +424,11 @@ export default class Editor extends EventEmitter {
       }
 
       this.remapNodeRefsInComponents(json.nodes, nodeDefs, uuidToIndexMap);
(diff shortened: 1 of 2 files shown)


DougReeder (approved):
LGTM

PR #1314, opened 2026-05-26
[Mozilla Branding Removal] Remove the last bits of Mozilla branding

Exairnous:
## What?

Removes Mozilla branding from the habitat scripts, some test files, images, videos, and a few other miscellaneous files.

## Why?

This is no longer a Mozilla project.

## Examples

N/A

## How to test

1. Run a search for the text "Mozilla" and see there are no uses of it other than for technical/credit reasons.
    * e.g. on GNU/Linux you can run this command to find any Mozilla references:`grep -r -i --exclude-dir={.git,.tersercache,dist,node_modules} "mozilla"`
2. Look through all the images and see there is no branding.
    * Running this command on GNU/Linux will list all of the images in the current directory and subdirectories `find . -type f -exec file -N -i -- {} + | sed -n 's!: image/[^:]*$!!p'`
3. Look through all the videos and see there is no branding.
    * Running this command on GNU/Linux will list all of the videos in the current directory and subdirectories `find . -type f -exec file -N -i -- {} + | sed -n 's!: video/[^:]*$!!p'`
4. Build a Docker image of the changes from this PR, deploy it to your Hubs instance, and see that everything still works as normal.

## Documentation of functionality

No functionality is changed, so no documentation update is needed.

## Limitations

This doesn't regenerate the AVA test snapshot or fix the tests (technically it introduces a new test error).  See the [Additional details or related context section](#additional-details-or-related-context) for why.  It also doesn't fix the linter errors as these are unrelated to the branding removal.

Old URLs to previously Mozilla run servers, but without Mozilla in the name (e.g. something.reticulum.io), were left as it isn't necessarily clear what they should be replaced with (or they function as a reminder/placeholder for things we'll be rehosting in the future, like the architecture kit) and they don't directly reference Mozilla.

## Alternative implementations considered

None.

## Open questions

None.

## Additional details or related context

The AVA test snapshot (test/integration/snapshots/Editor.test.js.snap)  will need to be regenerated with `yarn run update-test-snapshots` due to the changes in test/integration/snapshots/Editor.test.js.md, but that will need to wait until the architecture kit and crater.glb are rehosted in order to complete successfully.  Technically, this means one Mozilla reference is left (inside the snapshot), but it's only visible when running the tests and seeing it marked as an error.

This doesn't remove/modify non branding related Mozilla references like credits or code/dependencies as that would be inappropriate/break things.


DougReeder (approved):
LGTM

